PR #66, opened 2014-02-07
Add support for Darwin, at least when using MacPorts

mpdude:


mhaskel:
@mpdude thanks for the contribution, but since we haven't heard from you in a while I'm going to close this pull request. If this is functionality you'd still like to have added to the module, please update the tests and resbumit.

PR #69, opened 2014-03-14
Install npm on Ubuntu when not using PPA

emonty:
On Ubuntu, if we're not using the PPA, the npm package does need
to be explicitly installed. Modify the boolean logic around the
trap to only skip installing it on Ubuntu if we are, in fact,
also managing the PPA.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -91,7 +91,7 @@ class nodejs(
     require => Anchor['nodejs::repo']
   }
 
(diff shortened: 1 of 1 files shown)


mhaskel:
@emonty thanks for the contribution, but since we haven't heard from you in a while we're going to need to close this PR. @rchrd2 @atrepca @emonty if this is functionality you still want added to the module, please address the comments and resubmit.

Issue #51, opened 2013-10-07
Proxy config fails on ubuntu when using the managed repo


datr:
The proxy config command fails because it requires Package['npm'] but we don't install this from the managed repo as it's bundled.

juniorsysadmin:
Sorta fixed when using the npm defined type, but not sure if this would still raise its head when using the npm provider. Closing for now, but feel free to re-open.

PR #70, opened 2014-03-14
fix requires for proxy config when on ubuntu #51

wenlock:
Proxy config fails on ubuntu when using the managed repo #51


mhaskel:
Thanks for the contribution @wenlock

PR #77, opened 2014-05-20
Deploy from custom registry

javiplx:
Add install_options feature to enable the use of a custom registry. This avoids changes on .npmrc, and can be used on a per-package basis.

### DIFF
--- a/lib/puppet/provider/package/npm.rb
+++ b/lib/puppet/provider/package/npm.rb
@@ -70,10 +70,16 @@ Puppet::Type.type(:package).provide :npm, :parent => Puppet::Provider::Package d
       package = "#{resource[:name]}@#{resource[:ensure]}"
     end
 
+    extraopts = nil
+
     if resource[:source]
       npm('install', '--global', resource[:source])
     else
-      npm('install', '--global', package)
+      if extraopts.nil?
+        npm('install', '--global', package)
+      else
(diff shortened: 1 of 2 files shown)


mhaskel:
@javiplx @bdruth ping?

lib/puppet/provider/package/npm.rb
```diff
@@ -70,10 +70,24 @@ def install
       package = "#{resource[:name]}@#{resource[:ensure]}"
     end
 
+    extraopts = []
+
+    if resource[:install_options]
+      extraopts += resource[:install_options][0].collect do |k,v|
```
hunner: Why is `:install_options` an array, but then just access `[0]` ?

lib/puppet/provider/package/npm.rb
```diff
@@ -70,10 +70,24 @@ def install
...
+        if k == 'registry'
+          extraopts += [ "--registry" , v ]
```
hunner: Is this how the `install_options` for the package provider works? What about cases other than `registry`? Wouldn't it be better to just have keys like `{'--registry' => 'custom'}`?

javiplx:
This is nearly a year old PR, so I cannot anwer why I did what I did. The change worked on real world from the very beginning, although did never succeed the tests., and I've realized that currently it is not even mergeable.

PR #78, opened 2014-06-04
fix ensure => latest to work, and not rely on parsing free formatted text

teancom:
Previously, ensure => latest did not work, at least with the latest versions of npm. I switched over to using the JSON output so it will no longer break if npm changes output (which is what I think happened).


hunner:
I rebased to master in #115 to see if that fixes the errors, but got different errors.

PR #79, opened 2014-06-05
updated travis to current versions of puppet

jlambert121:


mhaskel:
@jlambert121 thanks for the contribution!

@igalic can't drop 2.7 without a major version bump...

igalic:
sadface

PR #89, opened 2014-09-28
changed the version from present to 0.10.32

wassimsalib:
"0.10.32-1chl1~precise1"


mhaskel:
@wassimsalib thanks for the contribution, but this looks like it's going to break the module everywhere but Ubuntu, so I'm going to close this pull request. In general we want the default values for parameters to make the most sense for all platforms. You can get this functionality by doing
```pupppet
class { 'nodejs':
  version => '0.10.32-1chl1~precise1',
}
```

PR #91, opened 2014-09-29
changed the version from present to 0.10.32

wassimsalib:


mhaskel:
@wassimsalib thanks for the contribution, but we don't want to change the default value for `$manage_repo`, so I'm going to close this PR. You can get this behavior in your environment by:
```puppet
class { 'nodejs':
  manage_repo => true,
}
```

PR #92, opened 2014-09-29
Add hiera support

dstockman:
Added hiera support, updated docs. Tested and in production on AWS Linux 2014.03.


mhaskel:
@dstockman Thanks for the contribution, but in general we like to keep all the hiera specific code in local roles and profiles, not in the modules themselves, so I'm going to close this PR.

@skpy looks like `$node_pkg` has been added to `init.pp`

PR #94, opened 2014-11-19
fixes for npm config proxy

CpuID:
- ensure noop when proxy is default (undef instead of empty string)
- fix the require package to be compatible with all OSes


mhaskel:
@CpuID Thanks for the contribution, but I believe this is fixed in #70 which was just merged, so I'm going to close this PR. If that doesn't work for you, please resubmit!

PR #99, opened 2014-12-04
If using standard ubuntu repos you need to install npm separately 

b4ldr:
Tested on ubuntu 14.04 however 12.04 appears to have a separate npm package as well


hunner:
Closed it favor of #103

PR #102, opened 2015-01-02
Fix for 3.7.3 (parser=future) Invalid relationship

x3dfxjunkie:
For Ubuntu, Gentoo and Archlinux, no npm package is defined.
```
Error: Could not retrieve catalog from remote server: Error 400 on SERVER: Invalid relationship: Exec[npm_proxy] { require => Package[npm] }, because Package[npm] doesn't seem to be in the catalog
```

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -169,13 +169,13 @@ class nodejs(
     }
   }
 
-  if $proxy {
-    exec { 'npm_proxy':
-      command => "npm config set proxy ${proxy}",
-      path    => $::path,
-      require => Package[$npm_package],
-    }
-  }
+  # if $proxy {
+  #   exec { 'npm_proxy':
+  #     command => "npm config set proxy ${proxy}",
+  #     path    => $::path,
+  #     require => Package[$npm_package],
+  #   }
+  # }
 
   if $dev_package and $dev_pkg {
     package { 'nodejs-dev':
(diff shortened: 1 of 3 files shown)


igalic:
i'm not sure if this test failure means that our code is now wrong, or our tests: https://travis-ci.org/puppetlabs/puppetlabs-nodejs/jobs/45721348#L401-L406

x3dfxjunkie:
That's the error we get, which is resolved by the patch I've submitted.  I'm unsure why this is just now coming up, but it clearly doesn't have an npm package to require in most cases.

igalic:
can you please fix [this test](https://github.com/puppetlabs/puppetlabs-nodejs/blob/master/spec/classes/nodejs_spec.rb#L253)?

x3dfxjunkie:
``` ruby
    it { should contain_exec('npm_proxy').with({
      'command' => 'npm config set proxy http://proxy.puppetlabs.lan:80/',
      'require' => 'Package[npm]',
    }) }
```

I'm unsure how best to fix this test.  The 'npm' package doesn't exist on all systems, so it shouldn't strictly require it by static name.

x3dfxjunkie:
Updated the test, and now it's passing.
https://travis-ci.org/puppetlabs/puppetlabs-nodejs/builds/45977450

hunner:
Closed in favor of #103

PR #103, opened 2015-01-06
(MODULES-1637) Major refactor

juniorsysadmin:
As mentioned in the Changelog and commit log, this Pull Request if commited will break a few things since major changes have been made to the API.

The code could probably be a bit better, so I have created this PR to get some feedback before I create a some acceptance tests.

@x3dfxjunkie - this PR potentially supercedes PR 102
@b4ldr - this PR potentially supercedes PR 99
@menssen - this PR potentially supercedes PR 96
@CpuID - this PR potentially supercedes PR 94
@aebm - this PR potentially supercedes PR 93
@wassimsalib - this might make your PR 89 redundant
@einyx - this PR potentially supercedes your PR 85
@jurgenweber, @brentax - this PR potentially supercedes PR 82
@jasperla - this PR potentially supercedes PR 80 - Are you perhaps able to check that it works nicely on FreeBSD?
@jlambert121 - this PR potentially supercedes PR 79 - Travis Puppet 2.7 tests have been removed, and I have attempted to use modulesync
@wenlock, @wknapik - this might make PR 70 redundant
@emonty, @rchrd2, @atrepca - this PR potentially supercedes PR 69
@mpdude - this PR potentially supercedes PR 66 - Are you perhaps able to confirm that the MacPorts default npm path is correct?

A few starting points of discussion:

1. It might be possible to modify this code to be backward-compatible with the current version, but it seems like a lot of work. If one follows the SemVer guidelines, breaking the API for 0.x releases is perfectly okay.

2. This code adds an explicit dependency on the treydock/gpg_key module for importing the YUM repository keys. I'm not sure if this is a something that is preferred for PuppetLabs modules in particular.

3. For whatever reason, Node.js often gets installed on a variety of operating systems, included 'bleeding edge' ones. The usual policy of failing hard when newer operating system versions are supported does
not seem to make sense for this module. Any strong objections?

4. The Travis CI Puppet 2.7 tests have been removed, because I errored out out with unusual messages on rspec tests, even though a real test on a Puppet 2.7 agent seemed to be successful.

5. These patches don't modify the npm (global) Puppet provider code, which can probably be left for a new PR. While this code forces the global config settings proxy and https-proxy to be applied first before any local npm packages are installed, it does not do something similar with the npm provider type.
There is a need for the npm_path parameter since Windows, Darwin and possibly others default to installing npm in various places. The npm defined types have been modified to accommodate this, but the provider has not.

6. NodeSource haven't released a binaries for Fedora 21, but I would think they are imminent.

7. I have updated the module version number to 0.7.0 in various places, and if this gets checked in the
tag on the master will need to be updated as well. This is rather optimistic, so I am happy to reverse these changes.


juniorsysadmin:
@skpy This incorporates your previous PR on parameterizing package names slightly differently, so you may wish to comment as well.

jasperla:
I'll check if this work on OpenBSD instead of FreeBSD as I've never done any work on this module for FreeBSD.

juniorsysadmin:
@jasperla My mistake. I have just pushed through another commit. Can you confirm that the default npm path is correct on OpenBSD?

README.md
```diff
@@ -1,84 +1,343 @@
...
-      node_pkg    => 'nodejs010',
-      npm_pkg     => 'nodejs010-npm',
-      dev_pkg     => 'nodejs010-devel',
-      dev_package => true,
-    }
+nodejs::npm is used for the local installation of npm packages:
 
-Debian users may want to use apt::pin to pin package installation priority on squeeze. See [puppetlabs-apt](https://github.com/puppetlabs/puppetlabs-apt) for more information.
+```puppet
+nodejs::npm { 'express|/opt/packages':
```
rnelson0: Describe the 'express|/opt/packages' string format.

README.md
```diff
@@ -1,84 +1,343 @@
...
+  ensure  => 'present'
+  tag     => 'tagname',
+}
+```
+
+Installing scoped packages:
+
+```puppet
+nodejs::npm { '@scope_name/express|/opt/packages':
```
rnelson0: Explain this string format as well.

README.md
```diff
@@ -1,84 +1,343 @@
...
+Installing packages from a local tarball:
+
+```puppet
+nodejs::npm { 'express|/opt/packages':
+  ensure  => 'present'
+  source  => '/local/repository/npm_packages/express.tgz'
+}
+```
+
+The nodejs::npm title consists of a package name separated by a '|', and
```
rnelson0: What about the '@scope_name' part? I think it would make sense to describe this before the examples.

manifests/params.pp
```diff
@@ -1,78 +1,203 @@
...
+        $repo_class                = '::nodejs::repo::nodesource'
+      }
+      elsif $::operatingsystemrelease =~ /^10.04$/ {
+        $manage_package_repo       = true
+        $nodejs_debug_package_name = 'nodejs-dbg'
+        $nodejs_dev_package_name   = undef
+        $nodejs_package_name       = 'nodejs'
+        $npm_package_ensure        = 'absent'
+        $npm_path                  = '/usr/bin/npm'
+        $npm_package_name          = undef
```
rnelson0: All other OS sections list $npm_package_name followed by $npm_path

x3dfxjunkie:
@juniorsysadmin, that looks like it would make my PR 102 irrelevant (not that I'm complaining).

wenlock:
@juniorsysadmin indeed that would do it.  +1 ty!

juniorsysadmin:
@jasperla - have fixed that in last commit
@rnelson0 - have added some documentation

rnelson0:
@juniorsysadmin I think the nodejs::npm docs at the top look better and it doesn't feel too verbose to me, just enough.

juniorsysadmin:
Rebased. Does anyone have any more feedback to add?

juniorsysadmin:
Rebased. Optimistic proposed version number changed to 0.8.0.

CHANGELOG.md
```diff
@@ -1,3 +1,26 @@
+##2015-01-21 0.8.0
```
mhaskel: That date's a little optimistic, as a release date. ;)

CHANGELOG.md
```diff
@@ -1,3 +1,26 @@
...
+###Summary
+
+This release performs major API changes and is not backwards-compatible with
+the previous release.
```
mhaskel: Yes, since this is still 0.x backwards incompat is fine, it's just best to be as explicit as possible with what's breaking in the changelog, so if you could add some information on what's backwards incompat that would be fantastic.

manifests/install.pp
```diff
@@ -0,0 +1,52 @@
+# == Class: nodejs::install
+class nodejs::install inherits nodejs {
```
mhaskel: I don't like that this is inheriting nodejs. It's fair to assume that `class { 'nodejs': }` will be declared before `nodejs::install`

manifests/repo/nodesource.pp
```diff
@@ -0,0 +1,95 @@
+# PRIVATE CLASS: Do not use directly
+class nodejs::repo::nodesource inherits nodejs {
```
mhaskel: Don't inherit from `nodejs`

manifests/repo/nodesource/apt.pp
```diff
@@ -0,0 +1,28 @@
+# PRIVATE CLASS: Do not use directly.
+class nodejs::repo::nodesource::apt inherits nodejs::repo::nodesource {
```
mhaskel: inherits.

manifests/repo/nodesource/yum.pp
```diff
@@ -0,0 +1,71 @@
+# PRIVATE CLASS: Do not use directly.
+class nodejs::repo::nodesource::yum inherits nodejs::repo::nodesource {
```
mhaskel: inherits

metadata.json
```diff
@@ -1,14 +1,94 @@
...
+      "version_requirement": ">=0.0.3 <1.0.0"
+    },
+    {
+      "name": "stahnma/epel",
+      "version_requirement": ">=1.0.0 <2.0.0"
+    },
+    {
+      "name": "treydock/gpg_key",
+      "version_requirement": ">=0.0.3 <1.0.0"
+    }
```
mhaskel: All of these platform-dependent requirements shouldn't be in the metadata, but rather documented in the README.md similar to https://github.com/puppetlabs/puppetlabs-rabbitmq#module-dependencies

metadata.json
```diff
@@ -1,14 +1,94 @@
...
+  ],
+  "requirements": [
+    {
+      "name": "pe",
+      "version_requirement": "3.x"
+    },
+    {
+      "name": "puppet",
+      "version_requirement": ">=2.7.20 < 4.0.0"
```
mhaskel: If this is backwards incompatible already can we just make this 3.x?

juniorsysadmin:
Rebased and some changes made based on feedback.

juniorsysadmin:
> NodeSource haven't released a binaries for Fedora 21, but I would think they are imminent.

NodeSource have released Fedora 21 binaries today on their repository.

igalic:
general comments: i remain unhappy with `node::npm { 'foo|/target/dir': ... }`
i would've thought that in a major refactor we would've changed this to `target => /target/dir`, but i guess the problem here is that we'll run into title uniqueness issues.

nibalizer:
I tend to agree with @igalic that the syntax used to hack in composite namvars is startling and my initial reaction to it is negative.

That being said, composite namevars have always been a 'maybe next year' so hacking it in with pipes in the node module, hey why not. The more I look at it, the more excited about the prospect I get.

I haven't been able to review the entire patch despite several attempts. This patch is super huge which is a problem for any reviewer. I hate to be "that guy" but could this be broken up into a few logical chunks? 

This seems like it is the best way forward for a real future for this module. @juniorsysadmin do you anticipate being able to take on primary ownership for this module going forward?

To respond directly to the conversation starters:

1) semver does state that you can YOLO on the entire 0.x lifetime. I think its great to keep it in 0.x so that we can massage this new syntax a bit. I think you should have no fear of breaking compat with the old version.

2)  A lot of modules do hardcode in the gpg keys. Rabbitmq does this. I don't have a preference, if a module makes this easier then I'm ok with it.

3) Not sure I understand this. Are you saying when an unrecognized operating system is detected it should fail, but you've wired it to not fail and keep on chugging? I suppose I think thats okay.

4) Puppet 2.7 should diaf. 

5) Largely the node::npm defined type is replacing the provider for the package resource. I think we can let the ruby provider lag in feature set until it's discovered what the utility of the new syntax for packages is figured out.

6) Sounds like this resolved itself.

7) Sounds good.

manifests/npm.pp
```diff
@@ -1,58 +1,88 @@
...
+    $package_string = "${npm_package}@${npm_tag}"
+  }
+  elsif $ensure == 'latest' {
+    $package_string = "${npm_package}@latest"
+  }
+  elsif $ensure == 'present' {
+    $package_string = $npm_package
+  }
+  elsif $ensure == 'absent' {
+    $package_string = $npm_package
```
igalic: these two branches can be collapsed.

manifests/npm.pp
```diff
@@ -1,58 +1,88 @@
...
-  } elsif $version {
-    $install_pkg = "${npm_pkg}@${version}"
-  } else {
-    $install_pkg = $npm_pkg
+    $package_string = $source
+  }
+  # install_check will always return false when tag is provided
+  # npm ls does not keep track of tags after install
+  elsif $npm_tag {
+    $package_string = "${npm_package}@${npm_tag}"
```
igalic: i'm not happy about the npm_tag parameter.
i'd rather we supply this via ensure, like we do everywhere else. (or, as we already allow, via the namevar.

juniorsysadmin:
>I haven't been able to review the entire patch despite several attempts. This patch is super huge which is a problem for any reviewer. I hate to be "that guy" but could this be broken up into a few logical chunks?

I could try and do this, but it's going to be really, really hard - the provider is the only code part that escaped unscathed, and that's because I don't have very strong knowledge of Puppet provider types.
It's a bit dishonest to call this a Pull Request - it's really just a fork masquerading as one. So you're better off just looking at the forked code rather than diffs.

>3) Not sure I understand this. Are you saying when an unrecognized operating system is detected it should fail, but you've wired it to not fail and keep on chugging? I suppose I think thats okay.

What my convoluted sentence was trying to say was:
- This module is not sticking with long term releases as the only 'supported' versions
- EOL operating system versions are 'not supported'
- Anything newer than a long term release it will chug away with a warning
That's the idea anyway.

>This seems like it is the best way forward for a real future for this module. @juniorsysadmin do you anticipate being able to take on primary ownership for this module going forward?

I can try - my knowledge of Puppet providers is not great though. The current version looks like an absolute nightmare for anyone who needs to merge PRs, I would think this version would make things a lot easier for whoever needs to merge PRs in future.

manifests/npm.pp
```diff
@@ -1,58 +1,88 @@
...
+  }
+  elsif $ensure == 'latest' {
+    $package_string = "${npm_package}@latest"
```
igalic: if that was the case, we could, again, collapse these two branches, i.e.: if ensure matches latest, or a version string, just append it.
in that case, it would make sense to make this the default, last case. i.e.: present, absent, etc.. have already been handled, now we just give up and decide it's a version number (or latest)

manifests/npm.pp
```diff
@@ -1,58 +1,88 @@
...
-    $validate = "${npm_dir}/node_modules/${npm_pkg}:${npm_pkg}@${version}"
-  } else {
-    $validate = "${npm_dir}/node_modules/${npm_pkg}"
+  $grep_command = $::osfamily ? {
+    'Windows' => 'cmd.exe /c findstr',
+    default   => 'grep',
   }
 
-  if $ensure == present {
+  if $ensure != 'absent' {
```
igalic: i would also collapse install and removal branch (the only difference being the direction of the grep and rm vs install), and install_options and uninstall_options into one.
that would make this whole define easier to read.

nibalizer:
@hlindberg courtesy ping so you can look at the '|' namevar syntax.

hlindberg:
@nibalizer From a language perspective, the title is any valid string and it is up to the type to do what it pleases with it. From a resource perspective it has to be unique. In practice, there are several types where a compound namevar is wanted and it would be a shame if it is done differently in different types.

My natural tendency would be to use URI syntax for this sort of thing - i.e a path within a domain. Protocol is not needed since that is really the type.

juniorsysadmin:
Based on feedback (especially about the namevar ugliness), the last commit:
- Allows an arbitary value for the namevar for the nodejs:npm defined type except if one wants to be lazy and not specify a package name
- Adds a user parameter to allow npm install to be run as a specific user

When ensure => latest is set, the installcheck will always fail and run npm install package@latest - I am not sure if there is much that can be done about this without using a Custom Type. Longer term, a better direction for this defined type is probably a wrapper which calls a Custom Type, similar to the what is done in puppetlabs-mysql etc.

juniorsysadmin:
Alignment issues in the RSpec files of the last commit have been fixed.

igalic:
:+1: for merge

eljohn1014:
When are you planning on merging this? we would like support for updated node

igalic:
@cyberious, @mhaskel: ping

nibalizer:
ya lets merge this +1

juniorsysadmin:
I have fixed two major issues in nodejs::npm - ensure => absent should now work properly, and default npm paths are now correct on Windows. (The commits have been squashed). I have also done a fair bit of testing of the npm defined type on Linux and Windows, so I'm reasonably satisfied that it works as advertised.

daenney:
> This code adds an explicit dependency on the treydock/gpg_key module for importing the YUM repository keys. I'm not sure if this is a something that is preferred for PuppetLabs modules in particular.

This shouldn't really be needed. All it takes is to drop one file, the key file, in the right directory.

nibalizer:
\o/

igalic:
it's done!

thank you everyone for reviewing, and thank you so much @juniorsysadmin!

mmarseglia:
Thank you @juniorsysadmin!

PR #106, opened 2015-01-21
Correct broken application of PR #70

theothertom:
PR #70 was merged in a way that causes a syntax error. I've moved the $npm_require into the case block, though it could also be in a separate if following the case.


mhaskel:
Thanks @theothertom

PR #109, opened 2015-02-05
added support for installing latest nodejs for Enterprise Linux from nod...

mmarseglia:
I needed to install the latest nodejs as the fedora repository is quite old.  I added boolean parameter to use nodesource.com repository for Enterprise Linux, added nodesource repository.  also added conditional around package npm install as npm is part of nodejs package from nodesource.


igalic:
this is handled in #103, which is due to be merged, real soon now.

mmarseglia:
Thank you for the update.

PR #110, opened 2015-02-11
Update init.pp

bribroder:
If node and npm are provided by the same package, don't try to require it twice


juniorsysadmin:
I'm beating my own drum here, but can you perhaps take a look at #103 ? That PR should have sensible defaults for this kind of thing. If not, feel free to make a comment. Thanks for your PR!

hunner:
This was solved with #103

PR #111, opened 2015-02-24
README.md

malnick:
Include that you need the dev package as a dependency when installing npm. Apt will fail to install npm if the dev package for nodejs is not included.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,10 +48,13 @@ To install Node.js and npm (using the NodeSource repository if possible):
 Install Node.js and npm using the native packages provided by the distribution:
 (Only applicable for Ubuntu 12.04/14.04 and Fedora operating systems).
 
+On Ubuntu 14+ the dev package needs to be installed as a dependency for npm:
+
 ```puppet
 class { '::nodejs':
   manage_package_repo => false,
   npm_package_ensure  => 'present',
+  nodejs_dev_package_ensure => 'present',
 }
 ```
 


igalic:
@malnick shouldn't we switch that on by default if npm is requested?

juniorsysadmin:
Okay, so at least on 14.04 the native nodejs-dev package is a dependency of the native npm package. I'll see if this is the case with 12.04 as well.

juniorsysadmin:
So to summarize to the best of my ability:
Ubuntu 12.04 and 14.04 native npm requires native nodejs-dev
RedHat families with a native npm package require native nodejs-devel
RedHat with EPEL npm requries EPEL nodejs-devel

npm is included with the nodejs package on Nodesource. (There is no nodejs-dev package in the Nodesource repo for Debian families)
With RedHat families, the Nodesource nodejs-devel package does not get pulled in when installing the Nodesource nodejs package - as expected.

I think changing the README docs is sufficient, but after the above points probably needs a little more.

juniorsysadmin:
@malnick Sent a PR to your fork. If you're happy with it I would suggest squashing up the commits and doing a slightly more descriptive commit message.

igalic:
imo, the people who package npm, pip, and other stuff for debian need some real world exposure to people using their packages. sorry, trying to use their packages.

i've never once had a case where i'd install pip (or npm) without having to install the -dev package, because one is useless without the other…

PR #112, opened 2015-02-26
Make sure nodejs::install is private

hunner:


manifests/install.pp
```diff
@@ -1,5 +1,8 @@
-# == Class: nodejs::install
+# PRIVATE CLASS: do not call directly
 class nodejs::install {
+  if $caller_module_name != $module_name {
+    fail("Use of private class ${name} by ${caller_module_name}")
+  }
```
igalic: can't we use `private()` here?

mhaskel: Nope. `private()` wasn't added until stdlib 4.4.0 or 4.5.0

PR #118, opened 2015-03-16
Improve Debian repository handling

paroga:


juniorsysadmin:
Thanks for this.

For 62c9823 are you able to add some docs to the README for repo_version, mentioning a URL if you have one (other than https://deb.nodesource.com/setup) and perhaps add some RSpec tests for this as well? This may be pretty important for anyone wanting to use 0.12.x as versions at https://deb.nodesource.com/node/ seem to have stopped at 0.10.x

In fact maybe leave 62c9823 for a separate PR and rename this one, as 0.12.x repo handling seems to be a bit in transition - https://deb.nodesource.com/setup does not match https://github.com/nodesource/distributions/blob/master/deb/setup for instance. https://github.com/nodesource/distributions/pull/68 may give an indication about what is going on.

On an unrelated note, I'm not seeing updated 0.12.x rpms or a change in the rpm repo URL to accommodate 0.10 to 0.12, so NodeSource might be doing some more repo changes soon.

paroga:
Is "repo_version" a good way to implement this? I'm not 100% happy with it.
Can you point me to some test, which test similar stuff?
nodesource provides packages for 0.12 via the "/node_0.12" directory on the same server, but as far as I've seen only for deb.

igalic:
this indicates for a node-(major)-version, than a repository version, so maybe that's a better name?

juniorsysadmin:
Not to pre-empt whether this module will in future handle io.js packages as well, but another idea is to call the variable repo_url_suffix and pass this all the way down to the repo class.

$repo_url_suffix => 'iojs_1.x' - for the NodeSource io.js repo: https://deb.nodesource.com/iojs_1.x
$repo_url_suffix => 'node_0.10' - for the NodeSource Node.js 0.10 repo (param default): https://deb.nodesource.com/node_0.10
$repo_url_suffix => 'node_0.12' - for the NodeSource Node.js 0.12 repo: https://deb.nodesource.com/node_0.12

juniorsysadmin:
@paroga RSpec example: https://gist.github.com/juniorsysadmin/ea973cb7b2635109189f

juniorsysadmin:
@paroga Have used this PR as the basis for #125 to try and speed things along.

PR #119, opened 2015-03-17
Fix gpg key checking warings after f588f26 in puppetlabs-apt

paroga:
Use the full fingerprint for the apt key to silence the warning.

### DIFF
--- a/manifests/repo/nodesource/apt.pp
+++ b/manifests/repo/nodesource/apt.pp
@@ -10,7 +10,7 @@ class nodejs::repo::nodesource::apt {
   if ($ensure == 'present') {
     apt::source { 'nodesource':
       include_src       => $enable_src,
-      key               => '1655A0AB68576280',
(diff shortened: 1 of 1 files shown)


juniorsysadmin:
Failing tests should pass when PR #120 is merged

xtreamwayz:
I have tried this fix and it breaks npm installation. When installing npm modules it throws: "Error: Command npm is missing".
Reverting to the short key installed npm and the modules, but gave the warning again obviously.

juniorsysadmin:
@xtreamwayz 
Verified as failing on puppetlabs/ubuntu-14.04-64-nocm 10.00 with puppetlabs-apt (v1.8.0) using Puppet 3.7.4. The problem appears to be related to dependency ordering and not specifically this PR though. (You're seeing this error because the native nodejs package got installed instead, which doesn't include npm)

juniorsysadmin:
@xtreamwayz The class containment issue should hopefully be fixed in #121

mhaskel:
Thanks @paroga !

PR #120, opened 2015-03-18
Pin RSpec gems

juniorsysadmin:
Keep this module consistent with other PL ones.

PR #121, opened 2015-03-19
Fix class containment

juniorsysadmin:
The `nodejs::repo::nodesource` class includes a few classes that are not contained properly, meaning for instance that the NodeSource repository might be created after the Node.js package is installed when the distribution comes with a native Node.js package. This PR is an attempt to fix this without using inheritance, making some anchor readability changes as well. Failing RSpec test should already be fixed in #120 .

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ##2015-02-27 0.8.0
 ###Backwards-incompatible changes
+- Puppet versions below 3.4.0 are no longer supported
 - Debian Squeeze and Fedora version 18 and below are explicitly no longer
   supported
 - Parameter naming changes to node_pkg, npm_pkg, dev_pkg, manage_repo,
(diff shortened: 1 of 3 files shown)


igalic:
*mutters something about [contain](http://docs.puppetlabs.com/references/3.7.latest/function.html#contain)*

juniorsysadmin:
I would prefer not to leave Puppet < 3.4.0 unsupported just yet

juniorsysadmin:
> I would prefer not to leave Puppet < 3.4.0 unsupported just yet

I've changed my mind on this - the last commit use contain and therefore drops support for any Puppet version less than 3.4.0. If there are no objections on this I will squash the commits shortly.

juniorsysadmin:
Squashed and rebased

bmjen:
Thanks @juniorsysadmin! Can you please rebase one more time to pick up the travis.yml updates from #127? Thanks!

juniorsysadmin:
Rebased

PR #122, opened 2015-03-20
Update apt key to full 40characters

exptom:
Latest version of puppetlabs/apt module shows warning on every puppet run if using a short key so this update includes the full 40 character key.


juniorsysadmin:
Thanks for this, but er I think #119 is a PR for this already :)

exptom:
Oh yes, didn't spot that!

PR #123, opened 2015-03-29
Update rspec-puppet tests for 2.0

juniorsysadmin:
Replace `subject` with `catalogue` for the raise_error matchers


igalic:
:+1:

PR #124, opened 2015-03-29
Fix yumrepo parameter typo

juniorsysadmin:
The yumrepo type sadly does not support the ensure parameter


mhaskel:
thanks @juniorsysadmin !

hunner:
Yumrepo got an ensure parameter in 3.5 http://docs.puppetlabs.com/references/3.5.latest/type.html#yumrepo-attribute-ensure

PR #125, opened 2015-04-06
Handle NodeSource Debian URL suffixes

juniorsysadmin:
Based on PR #118 but adds RSpec tests and allows those who want bleeding fingers to install io.js with:

```puppet
class { '::nodejs':
  package_name => 'iojs',
  repo_url_suffix => 'iojs_1.x',
}
```

This PR still defaults to using Node.js 0.10.x as the latest version on Debian platforms rather than 0.12.x for the moment. (Note that NodeSource does not host a 0.12.x version for Lucid)

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,16 @@ On Windows the packages are installed via Chocolatey.
 To install Node.js and npm (using the NodeSource repository if possible):
 
 ```puppet
-    class { 'nodejs': }
+class { 'nodejs': }
+```
+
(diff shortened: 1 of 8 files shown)


juniorsysadmin:
The failing test is due to Travis defaulting to `sudo: false` and failing on an old npm provider test. This will need to be fixed in future PR by (preferably) fixing the provider test, or setting `sudo: true` in the travis.yml

igalic:
cool.

PR #126, opened 2015-04-23
Document that puppetlabs-apt versions less than 2.x are required

juniorsysadmin:
puppetlabs-apt 2.x changes apt::source parameters substantially. This patch makes it clear that this module will break if puppetlabs-apt 2.x is used currently.

The idea is that this module targets puppetlabs-apt 1.x up to the next tagged release (0.8.x), then this module gets modified to target puppetlabs-apt 2.x-only after this point (eg. puppetlabs-nodejs 0.9.x targets puppetlabs-apt 2.x)

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -1,7 +1,9 @@
 fixtures:
   repositories:
     "stdlib": "git://github.com/puppetlabs/puppetlabs-stdlib.git"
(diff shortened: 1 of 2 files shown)


nibalizer:
Apt is an implicit requirement in metadata.json. I am okay with this change.

PR #130, opened 2015-05-03
Fix stdlib and Puppet version requirements in metadata.json

juniorsysadmin:
Official stdlib docs state that stdlib 2.x is not compatible with Puppet 3.x.
We require Puppet versions 3.4.0 or greater due to #121

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -62,10 +62,10 @@
     },
     {
       "name": "puppet",
-      "version_requirement": ">=3.0.0 < 4.0.0"
+      "version_requirement": ">=3.4.0 < 4.0.0"
     }
   ],
   "dependencies": [
-    {"name":"puppetlabs/stdlib","version_requirement":">=2.2.1 <5.0.0"}
+    {"name":"puppetlabs/stdlib","version_requirement":">=3.0.0 <5.0.0"}
   ]
 }


igalic:
what's up with the test failures?
```
 1) nodejs when run on Debian release 7.0 with legacy_debian_symlinks set to true the file resource /usr/bin/node should be in the catalog
     Failure/Error: should contain_file('/usr/bin/node')
     Puppet::Error:
       undefined method `ref' for nil:NilClass on node testing-worker-linux-f25aadab-1-20168-linux-8-61003744
     # /home/travis/.rvm/gems/ruby-1.9.3-p551/gems/puppet-3.4.3/lib/puppet/resource/catalog.rb:527:in `block in to_catalog'
```

juniorsysadmin:
The test were failing for some reason only on 3.4.0 - I have no idea why. Since the travis.yml has now changed on master, I have rebased and tests on the new Travis CI matrix are now all green.

juniorsysadmin:
Before this was moved across, tests were okay. This should probably be merged in before the next release is cut. Any objections?

ffrank:
Can you make this into a new PR so that Travis can greenlight it once more? Call me a paranoid nitpick :)

nibalizer:
@ffrank why does it need a new PR? I see greens from travis and here: http://planck.nibalizer.com:5000/modules/puppet-community/puppet-nodejs

ffrank:
Rebase apparently did the trick? I'm happy either way :+1:

PR #135, opened 2015-05-14
Install NodeJS (and NPM) before any NPM packages

tkocheran-chromeriver:
This simple resource ordering statement makes sure that all NPM packages are installed _after_ the whole NodeJS class finishes its work.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -54,4 +54,6 @@ class nodejs(
     Class['::nodejs::install'] ->
     anchor { '::nodejs::end': }
   }
+  
+  Class['::nodejs'] -> Package <| provider == 'npm' |>
 }


juniorsysadmin:
@tkocheran-chromeriver Out of interest can you provide some example code that would reproduce this problem? I'm surprised that this is happening, but then again I haven't written some beaker acceptance tests yet so...

tkocheran-chromeriver:
```
class nodepackages {
    include ::node

    package { 'bower':
        provider => 'npm'
    }
}
```

Something like that ought to break it. What should get installed first? I don't think that Puppet automagically determines that it should install the provider (ie Class[node]) before the package with that provider. It _could_ be that smart, but my experiences serve otherwise.

ffrank:
Clever idea, but I'm uncomfortable with the possible side effect that virtual packages are getting realized without intent nor any means of circumvention on the user's side.

juniorsysadmin:
I may be wrong, but shouldn't [pluginsync] (https://docs.puppetlabs.com/guides/plugins_in_modules.html) already do this automatically?

ffrank:
Only half of it. The **provider** will be there, yes, but it won't become **suitable** on the system until the `node` class has done its work.

nibalizer:
Right but don't puppet resources defer if they provider isn't yet suitable? I agree with Frank that using a collector like this is not something a library should do because it will realize the virtual resources.

tkocheran-chromeriver:
@ffrank What are virtual packages and why is this a bad practice?

As a user, with the following class:

```
class nodepackages {
    include ::node

    package { 'bower':
        provider => 'npm'
    }
}
```

I'd expect that since Class[node] provides the 'npm' provider, Puppet would install everything from that class first. A collector would ensure that this happens. What other options do I have? Should I as the user have to write a collector which should be provided by the provider?

ffrank:
@nibalizer good point about defered resources. @tkocheran-chromeriver can you provide an example that breaks without this patch?

Virtual resources:

    class nodebase {
        include ::node

        @package { 'bower': provider => 'npm' }
    }

Here the package is declared as *virtual* thanks to the leading `@`. This way, the different profiles that use the `nodebase` class can choose wether the `bower` package should be realized or not. One way of realizing it is the collector. If it's part of the `node` module though, this does not work because the virtual package is *always* realized.

nibalizer:
Felix, I'm realizing that I don't have a good blog post or documentation to this exact issue. Do you know of anything or should I go ahead and write it up?

ffrank:
Could you? That would be splendid!

igalic:
@ffrank, @nibalizer: also: refreshonly

igalic:
I HAVE AN IDEA. HOW ABOUT MY PUPPET IN ANGER TALK BECOMES A LITTLE BOOKLET ON WORSE PRACTICES, ANTIPATTERNS, AND THE PAIN OF DEBUGGING

ffrank:
We're digressing, but yes, antipattern talks are always great.

Spencer, I'm not sure anymore what you are referring to. Will you write about resources defering themselves and how that works? I'll gladly do a bit about `refresonly`.

@tkocheran-chromeriver how would you feel about making this a README change, advising users to add the `before` override in their own manifests?

tkocheran-chromeriver:
@igalic I ALSO HAVE ANGER IN PUPPET.

@ffrank That's fine. I probably just need to learn about all the new features in Puppet. Virtual packages, Hiera, all the new Puppet 3 stuff is still not known to me. I still just write hierarchies of classes, classes for roles, classes for profiles, defined types in Puppet lang, etc. This is probably largely _my_ problem for not knowing more about all these new features.

ffrank:
Cool. What I also meant was - would you care to change your branch accordingly? I'd feel good about merging that then.

tkocheran-chromeriver:
Is there anything to be contributed? My understanding thus far is that my pull request is unnecessary.

ffrank:
The manifest statement is useful for users, but shouldn't be part of the module code.

If you could at it to `README.md` as an advice to users, that would be most helpful.

PR #144, opened 2015-05-26
Require treydock/gpg_key in metadata.json

everestx:


nibalizer:
The readme calls out dependencies on this, the apt module, epel, gentoo-portage, and chocolatey. These are all I believe implicitly called implicit dependencies. If we're going to promote implicit dependencies to real dependencies we should be consistent about that.

everestx:
Yes... I saw those too. If that's the case, then this can be ignored. I half expected to get that response, but I wasn't sure if the jump to version 1.0.0 just missed it.

nibalizer:
I'd be interested to see what others think about implicit dependencies.

juniorsysadmin:
I asked around about this earlier and Puppetlabs + others indicated that implicit dependencies should not be in the metadata.json.
By the way with Puppet 3.6 (I think) and higher, there is no need for treydock/gpg_key if there are also some key-related additions done to the yumrepo resource as well. But that is a potential PR for another day.

nibalizer:
@juniorsysadmin of course implicit dependencies won't be in metadata.json, that is what makes them implicit dependencies. The question, in my mind, is if having implicit dependencies is a good idea.

juniorsysadmin:
The implicit dependencies are spelled out in the README, so...

nibalizer:
Closing this, as the status quo seems to be okay with people. Feel free to re-open if you want to continue the conversation. @everestx thanks for the patch.

everestx:
Np. Thanks for the feedback everyone.

PR #150, opened 2015-07-18
Update .travis.yml

juniorsysadmin:
Test against Puppet 4 and attempt to fix a deploy issue where multiple attempts to push to the Forge are done, rather than once.


juniorsysadmin:
Any comments on this before I merge?

nibalizer:
Interesting that we deploy on 2.1.5. I think on most other modules we deploy on 1.9.3.

Issue #140, opened 2015-05-20
Fix deprecated use of should syntax in provider RSpec test


juniorsysadmin:
One of the provider tests uses a deprecated `should` form in its RSpec test. This should be replaced with the supported equivalent.

```
Using `should` from rspec-expectations' old `:should` syntax without explicitly enabling the syntax is deprecated.
Use the new `:expect` syntax or explicitly enable `:should` instead.
Called from ../puppet-nodejs/spec/unit/puppet/provider/package/npm_spec.rb:18:in `block (2 levels) in it_should_respond_to'.
```

igalic:
should we just run [transpec](http://yujinakayama.me/transpec/)?

PR #151, opened 2015-07-18
RSpec tests: Convert should to expect using transpec

juniorsysadmin:
Since the should syntax in RSpec is deprecated, convert to the new expect syntax.


juniorsysadmin:
Fixes #140

PR #152, opened 2015-07-18
Remove soft dependency on treydock/gpg_key

juniorsysadmin:
Remove soft dependency on treydock/gpg_key


juniorsysadmin:
Needs to be verified before merging.

igalic:
do we not have acceptance tests for that :P

tsde:
Just tested your changes on centos 6 and 7 flawlessly. Would be great to see this PR merged ;)

igalic:
we'll just merge it, and let it sort out :P

PR #155, opened 2015-07-28
fixing amazon default settings

webratz:
Amazon LInux (at least on Version 2015.03) reports its own osfamily as RedHat therefore the old code causes the manifest to fail.
This PR fixes the detection.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -98,6 +98,16 @@ class nodejs::params {
         $npm_path                  = '/usr/bin/npm'
         $repo_class                = '::nodejs::repo::nodesource'
       }
+      elsif ($::operatingsystem == 'Amazon') {
+        $manage_package_repo       = true
+        $nodejs_debug_package_name = 'nodejs-debuginfo'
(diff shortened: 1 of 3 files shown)


juniorsysadmin:
Good catch. Perhaps leave the previous code in there (with a comment explaining that this is for older versions of Amazon Linux (<2015.03)). Also, if you are able to, can you add the relevant unit tests as well?

webratz:
readded the test case, fixed also nodesource rpm for amazon linux.
updated the tests which work now on my local machine

igalic:
@webratz would you mind squashing the failing tests, please?

webratz:
squashed alltogether into first commit

igalic:
duh.
that was a bit too quickly fired. :(

webratz:
that looks indeed weird. i don't know how it happened but i guess we can't revert it easily

igalic:
@webratz no worries.
next time, do it like so: https://github.com/puppetlabs/puppetlabs-apache/pull/1167#issuecomment-126346474

PR #156, opened 2015-07-30
Fix failing provider RSpec test when running on docker

juniorsysadmin:
This hack mocks an exit success so that the test does not fail when
run on docker.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 ---
-sudo: true
+sudo: false
 language: ruby
 bundler_args: --without system_tests
 script: "bundle exec rake validate && bundle exec rake lint && bundle exec rake spec SPEC_OPTS='--format documentation'"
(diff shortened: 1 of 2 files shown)


juniorsysadmin:
For whatever reason `$CHILD_STATUS.success` does not evaluate to `true` when this test is run on TravisCI.

`
Puppet.debug("Warning: npm list --json exited with code #{$CHILD_STATUS.exitstatus}") unless $CHILD_STATUS.success?
`

This is a hack because there's no way to reproduce a way that npm will produce non-JSON output with `npm list -g --json` without a resulting exit failure - this is a unit test just for coverage that doesn't reflect normal output. Ideally, we probably would want to test the existence of both Puppet.debug statements being the output, but I couldn't think of a way to do this.

igalic:
thanks @juniorsysadmin!

PR #157, opened 2015-08-07
Exodusftw qualify global config var

exodusftw:
modified behavior of global_config_entry type by including "onlyif" attribute on exec to ensure config variables are only added/removed when set to absent or whenever present value doesn't match desired value

### DIFF
--- a/README.md
+++ b/README.md
@@ -283,7 +283,7 @@ nodejs::npm { 'remove all express packages':
 
 ### nodejs::npm::global_config_entry
 
-nodejs::npm::global_config_entry can be used to set global npm configuration settings.
+nodejs::npm::global_config_entry can be used to set/remove global npm configuration settings.
 
 Examples:
 
--- a/manifests/npm/global_config_entry.pp
+++ b/manifests/npm/global_config_entry.pp
@@ -8,12 +8,31 @@ define nodejs::npm::global_config_entry (
 
   validate_re($ensure, '^(present|absent)$', "${module_name}::npm::global_config_entry : Ensure parameter must be present or absent")
 
-  $command = $ensure ? {
-    'absent' => "config delete ${config_setting}",
-    default  => "config set ${config_setting} ${value} --global",
+  case $ensure {
+    'absent': {
+      $command        = "config delete ${config_setting} --global"
+      $onlyif_command = $::osfamily ? {
+        'Windows' => "CMD /C ${npm_path} get --global| FINDSTR /B \"${config_setting}\"",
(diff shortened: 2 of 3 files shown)


juniorsysadmin:
Just a few things:
- It's better if you add --global as part of ${command} to avoid confusion
- Use ${npm_path} rather than putting the whole path
..which leads to...
- Windows support. I attempted to support Windows in npm.pp with the use of alternatives for grep. I'm not sure if there is a Windows alternative to `tr` though. If you can't come up with changes that might allow this defined type to work on Windows, please put a comment in the appropriate section of the README and perhaps a code comment as well

exodusftw:
I worked through some issues on the windows onlyif components and it should now work as expected to be idempotent and not run the global config exec when not required.

I also ensured that the `${npm_path}` variable and and `--global` entries were added.

Please review.

juniorsysadmin:
If you do: ```$cmd_exe_path = $nodejs::cmd_exe_path,``` in the defined type, then you've already got the full path of cmd and can avoid the provider in the exec. I'm not sure if that can make things less ugly though. Anyhow, if you've tested on a Windows machine and am happy with this, just squash the commits and we'll merge it.

igalic:
@exodusftw i think this will beed a proper [rebase](http://git-scm.com/book/en/Git-Branching-Rebasing)
@juniorsysadmin is this now addressing all of your concerns?

juniorsysadmin:
Okay with merging after a rebase, but would prefer the use of ```$cmd_exe_path``` to slightly simplify things though. I know, I know, it should be a provider....

exodusftw:
I will update to use the cmd path var later this evening.


Sent by Outlook for Android





On Tue, Sep 22, 2015 at 5:08 AM -0700, "Ben S" <notifications@github.com> wrote:
Okay with merging after a rebase, but would prefer the use of ```$cmd_exe_path``` to slightly simplify things though. I know, I know, it should be a provider....

---

exodusftw:
Updated to use `$cmd_exe_path` - please review

igalic:
this looks quite good now, @exodusftw
would you mind doing a [squash](http://www.git-scm.com/book/en/Git-Tools-Rewriting-History#Squashing-Commits) ?

exodusftw:
Had some weirdness on the squash - but it should be sorted out - let me know if you want anymore cleanup.

juniorsysadmin:
:+1:

igalic:
@exodusftw rebased without the merge commit for you :)
thank you very much!

PR #161, opened 2015-08-17
Allowed home environment variable to be set

mootpt:
This was fixed in the package provider, but never fixed in the exec the nodejs::npm class was utilizing for installations:

Heres a bit of background:

Node-gyp expects either HOME or USERPROFILE to be defined in the environment. If it is not set, installation will fail.

```
Error: node-gyp requires that the user's home directory is specified in either of the environmental variables HOME or USERPROFILE
```

### DIFF
--- a/spec/defines/nodejs_npm_spec.rb
+++ b/spec/defines/nodejs_npm_spec.rb
@@ -47,19 +47,25 @@ describe 'nodejs::npm', :type => :define do
       it 'the command should be npm install express' do
         should contain_exec('npm_install_express').with({
           'command' => '/usr/bin/npm install express ',
-         })
+        })
       end
 
       it 'the install_check should grep for /home/npm/packages/node_modules/express' do
         should contain_exec('npm_install_express').with({
           'unless'  => "/usr/bin/npm ls --long --parseable | grep \"/home/npm/packages/node_modules/express\"",
-         })
+        })
       end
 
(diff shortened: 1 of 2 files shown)


mootpt:
This is related to: https://github.com/puppet-community/puppet-nodejs/pull/35

mootpt:
thanks @hunner

spec/defines/nodejs_npm_spec.rb
```diff
@@ -61,6 +61,12 @@
           'cwd'  => '/home/npm/packages',
          })
       end
+      
+      it 'the environment variable HOME should be /root' do
+        is_expected.to contain_exec('npm_install_express').with({
+          'environment'  => 'HOME=/root',
+         })
```
nibalizer: It this whitespace off?

mootpt: appears to be normal

nibalizer: There is an extra space before the } causing it not to line up.

spec/defines/nodejs_npm_spec.rb
```diff
@@ -61,6 +61,12 @@
           'cwd'  => '/home/npm/packages',
          })
       end
+
```
nibalizer: There are four extra spaces here.

PR #169, opened 2015-09-30
Ensure "apt update" after adding/deleting sources

jordiclariana:
Ensure "apt update" is executed after adding/deleting sources so expected nodejs version is installed.


igalic:
this is already the default in 1.8.x: https://github.com/puppetlabs/puppetlabs-apt/blob/1.8.x/manifests/source.pp#L48 and in 2.0.x https://github.com/puppetlabs/puppetlabs-apt/blob/master/manifests/setting.pp#L49-L53

jordiclariana:
True, my mistake. I had some problems with dependencies and I thought it was because of this. Clearly not.

Thanks anyway.

Issue #171, opened 2015-10-02
Nodejs package is installed before apt is updated


atrepca:
**OS:** Ubuntu 14.04.2 LTS
**Puppet version:** 3.7.2-1puppetlabs1
**Nodejs module version:** 0.8.0

**Puppet code snippet:**
```
  $npm_packages = [ 'bower','grunt','grunt-cli', ]
  package { $npm_packages:
    ensure   => present,
    provider => 'npm',
  }
  class { 'nodejs':
    repo_url_suffix       => 'node_0.12',
    nodejs_package_ensure => '0.12.7-1nodesource1~trusty1',
    before                => Package[$npm_packages],
  }
```
**Puppet run output:**
```
Info: Retrieving plugin
Info: Caching catalog for REMOVED
Info: Applying configuration version '1443784901'
Notice: /Stage[main]/Nodejs::Repo::Nodesource/Nodejs::Repo::Nodesource::Apt/Apt::Source[nodesource]/Apt::Setting[list-nodesource]/File[/etc/apt/sources.list.d/nodesource.list]/ensure: created
Info: /Stage[main]/Nodejs::Repo::Nodesource/Nodejs::Repo::Nodesource::Apt/Apt::Source[nodesource]/Apt::Setting[list-nodesource]/File[/etc/apt/sources.list.d/nodesource.list]: Scheduling refresh of Class[Apt::Update]
Error: Could not update: Execution of '/usr/bin/apt-get -q -y -o DPkg::Options::=--force-confold --force-yes install nodejs=0.12.7-1nodesource1~trusty1' returned 100: Reading package lists...
Building dependency tree...
Reading state information...
The following NEW packages will be installed:
  nodejs
0 upgraded, 1 newly installed, 0 to remove and 192 not upgraded.
E: Can't find a source to download version '0.12.7-1nodesource1~trusty1' of 'nodejs:amd64'
Wrapped exception:
Execution of '/usr/bin/apt-get -q -y -o DPkg::Options::=--force-confold --force-yes install nodejs=0.12.7-1nodesource1~trusty1' returned 100: Reading package lists...
Building dependency tree...
Reading state information...
The following NEW packages will be installed:
  nodejs
0 upgraded, 1 newly installed, 0 to remove and 192 not upgraded.
E: Can't find a source to download version '0.12.7-1nodesource1~trusty1' of 'nodejs:amd64'
[...]

PR #173, opened 2015-11-04
Added puppetlab-apt dependency

abednarik:
README says

This modules uses puppetlabs-apt for the management of the NodeSource repository. If using an operating system of the Debian-based family, you will need to ensure that puppetlabs-apt version 2.x is installed.

Related to h#172 and probably #171

### DIFF
--- a/manifests/repo/nodesource/apt.pp
+++ b/manifests/repo/nodesource/apt.pp
@@ -27,6 +27,7 @@ class nodejs::repo::nodesource::apt {
         Package['apt-transport-https'],
         Package['ca-certificates'],
       ],
+      notify   => Exec['apt_update'],
     }
   }
 
--- a/metadata.json
+++ b/metadata.json
@@ -67,6 +67,7 @@
   ],
   "dependencies": [
     {"name":"puppetlabs/stdlib","version_requirement":">=4.1.0 <5.0.0"},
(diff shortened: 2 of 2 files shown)


manifests/repo/nodesource/apt.pp
```diff
@@ -27,6 +27,7 @@
         Package['apt-transport-https'],
         Package['ca-certificates'],
       ],
+      notify   => Exec['apt_update'],
```
igalic: this should happen automatically.

metadata.json
```diff
@@ -67,6 +67,7 @@
   ],
   "dependencies": [
     {"name":"puppetlabs/stdlib","version_requirement":">=4.1.0 <5.0.0"},
+    {"name":"puppetlabs/apt","version_requirement":">=2.1.0 <3.0.0"},
```
igalic: since it's an optional dependency, we didn't list it.

metadata.json
```diff
@@ -67,6 +67,7 @@
...
     {"name":"treydock/gpg_key","version_requirement":">=0.0.3 <1.0.0"}
```
igalic: which makes me wonder why we listed this one…

ffrank: That should be OS agnostic, so I guess it's OK to pull it automagically?

PR #176, opened 2015-11-17
Fix repo_url_suffix regex validation

tsde:
Hi guys,

This PR is linked to #174 
I made a small mistake in regex checks for repo_url_suffix. This could lead to **$repo_url_suffix** not raising an error when wrongly set to **0.10.0** for example (the **0.10** is matched only so validate_re is not failing). This PR fixes that.

Sorry for that ;)

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -26,7 +26,7 @@ class nodejs(
 
   # Validate repo_url_suffix. Not every versions of NodeJS are available
   # for all distros at Nodesource. We need to check that.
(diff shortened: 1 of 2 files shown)


igalic:
@tsde would you mind adding a test to catch that assumption?

tsde:
@igalic I've added some test. And, bonus, also included tests on repo_url_suffix for redhat-based OS. Previously, there were no such tests for these distros.

Let me know if everything is fine

igalic:
nice! thank you @tsde

PR #177, opened 2015-11-18
Update the puppet version requirement since tests and implementations…

pskrz:
… are successful against Puppet 4 on Ubuntu 14.04 agent (puppetserver-2.1.1/puppet-agent-1.2.1)

Fixes https://github.com/puppet-community/puppet-nodejs/issues/175

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -62,7 +62,7 @@
     },
     {
       "name": "puppet",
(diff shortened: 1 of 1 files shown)


igalic:
thanks @pskrz!

PR #183, opened 2015-12-15
Updates README with spaceship collector advice

petems:
Superceeds https://github.com/puppet-community/puppet-nodejs/pull/135

Since puppetlabs-apt has a similar caveat, best to keep it out of code IMO: https://github.com/puppetlabs/puppetlabs-apt#adding-new-sources-or-ppas

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,6 +120,30 @@ package { 'mime':
 }
 ```
 
+#### Caveat using the npm package provider
+
+Note: The npm provider won't be avaliable until the server has been marked as having the `npm` command avaliable.
+
+So the following code will not work:
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -120,6 +120,30 @@ package { 'mime':
 }
 ```
 
+#### Caveat using the npm package provider
+
+Note: The npm provider won't be avaliable until the server has been marked as having the `npm` command avaliable.
+
+So the following code will not work:
+
+```
```
igalic: → '```puppet'

petems: fixed :+1:

PR #188, opened 2016-01-07
Update from voxpupuli modulesync_config

igalic:


juniorsysadmin:
Rubocop fixes will be done in a subsequent PR

PR #190, opened 2016-01-07
WIP: Attempt to fix failng Rubocop lint checks

juniorsysadmin:


lib/puppet/provider/package/npm.rb
```diff
@@ -51,8 +51,8 @@ def query
   end
 
   def latest
-    if /#{resource[:name]}@([\d\.]+)/ =~ npm('outdated', '--global',  resource[:name])
-      @latest = $1
+    if /#{resource[:name]}@([\d\.]+)/ =~ npm('outdated', '--global', resource[:name])
+      @latest = Regexp.last_match(1)
     else
       @property_hash[:ensure] unless @property_hash[:ensure].is_a? Symbol
     end
```
igalic: this whole function feels very unruby… not sure ow to make it better
or simplify it for that matter
perhaps [split](http://ruby-doc.org/core-2.3.0/String.html#method-i-split) is all we need

so, basically,
```ruby
def latest
  outdated = npm('outdated', '--global', resource[:name]).split('@', 1) # no idea what i'm doing
  return outdated if outdated =~ /^(\d\.)+$/ # n.b.: this doesn't match builds, alpha's rcs, etc…
  return @property_hash[:ensure] unless @property_hash[:ensure].is_a? Symbol # why? we could just to_s
  # cuz otherwise this returns nil
end
```

spec/defines/nodejs_npm_spec.rb
```diff
@@ -23,348 +21,317 @@
...
+        }
+      end
 
       it 'the npm install command should run under user foo' do
-        is_expected.to contain_exec('npm_install_express').with({
-          'command' => '/usr/bin/npm install express ',
-          'user'    => 'foo',
-         })
+        is_expected.to contain_exec('npm_install_express').with('command' => '/usr/bin/npm install express ',
+                                                                'user'    => 'foo',)
```
igalic: dear rubocop,

this looks stupid.

bye,
i

spec/unit/puppet/provider/package/npm_spec.rb
```diff
@@ -22,48 +22,47 @@ def self.it_should_respond_to(*actions)
...
       @provider.class.instance_variable_set(:@npmlist, nil)
     end
 
-    it "should return a list of npm packages installed globally" do
+    it 'should return a list of npm packages installed globally' do
       @provider.class.expects(:execute).with(['/usr/local/bin/npm', 'list', '--json', '--global'], anything).returns(my_fixture_read('npm_global'))
-      expect(@provider.class.instances.map {|p| p.properties}.sort_by{|res| res[:name]}).to eq([
-        {:ensure => '2.5.9' , :provider => 'npm', :name => 'express'},
-        {:ensure => '1.1.15', :provider => 'npm', :name => 'npm'    },
+      expect(@provider.class.instances.map(&:properties).sort_by { |res| res[:name] }).to eq([
```
igalic: this is a bit more readable

spec/unit/puppet/provider/package/npm_spec.rb
```diff
@@ -22,48 +22,47 @@ def self.it_should_respond_to(*actions)
...
     end
 
-    it "should log and continue if the list command has a non-zero exit code" do
+    it 'should log and continue if the list command has a non-zero exit code' do
       @provider.class.expects(:execute).with(['/usr/local/bin/npm', 'list', '--json', '--global'], anything).returns(my_fixture_read('npm_global'))
       Process::Status.any_instance.expects(:success?).returns(false)
       Process::Status.any_instance.expects(:exitstatus).returns(123)
       Puppet.expects(:debug).with(regexp_matches(/123/))
-      expect(@provider.class.instances.map {|p| p.properties}).not_to eq([])
+      expect(@provider.class.instances.map(&:properties)).not_to eq([])
```
igalic: yup, again. more readable

igalic:
merging this for now, thank you @juniorsysadmin!

igalic:
i'll do the rest

spec/defines/nodejs_npm_spec.rb
rnelson0: May I suggest:
````
$ cat >> .rubocop.yml
Style/BlockDelimiters:
  Enabled: False

Style/MultilineBlockLayout:
  Enabled: False
````

igalic: @rnelson0 patches welcome

rnelson0: If you're interested in going back to the previous format, here's the diff required. You do not need to Modulesync as those block checks actually aren't related (but still good IMO):
````
$ git diff
diff --git a/spec/defines/nodejs_npm_spec.rb b/spec/defines/nodejs_npm_spec.rb
index 88a93d2..b4e27f2 100644
--- a/spec/defines/nodejs_npm_spec.rb
+++ b/spec/defines/nodejs_npm_spec.rb
@@ -30,8 +30,10 @@ describe 'nodejs::npm', :type => :define do
       end

       it 'the npm install command should run under user foo' do
-        is_expected.to contain_exec('npm_install_express').with('command' => '/usr/bin/npm install express ',
-                                                                'user'    => 'foo',)
+        is_expected.to contain_exec('npm_install_express').with(
+          'command' => '/usr/bin/npm install express ',
+          'user'    => 'foo',
+         )
       end
     end
````

daenney: I vastly prefer the multiline-version too. Much easier to read and no annoying scroll off.

PR #195, opened 2016-01-08
Change npm_package_name from "undef" to "false" for better comparison.

### LINKED ISSUE #160: Use npm to install a specific npm version
Hey all, 

I recently downloaded the nodejs module and I think it's really well done. I have run into a small feature request though. I'm using Ubuntu and the NodeSource packages, which (I believe) puts the npm binary in the same package as nodejs, instead of having 2 separate packages like the base Ubuntu packages. 

I would like to be able to use npm to upgrade npm to a specific version. When I set $npm_package_ensure to be a specific version it errors out though because it's trying to install that version via the default provider instead of via the npm provider. 

I'd love to add another variable that would allow you to override the provider for npm (something like $npm_provider).  Are there any downsides to doing this?  Is there a better way to get a specific version of npm using npm?  I'd love any feedback that the community would like to share.

Thanx!

x3dfxjunkie:
Fixes #160 

Before:
```
# puppet apply -v  -e 'class { "nodejs": npm_package_name => undef,} package { "npm": ensure => "2.14.14", provider => "npm",}'
Info: Loading facts
Info: Loading facts
Error: Duplicate declaration: Package[npm] is already declared in file /etc/puppet/modules/nodejs/manifests/install.pp:52; cannot redeclare at line 1 on node ***
Error: Duplicate declaration: Package[npm] is already declared in file /etc/puppet/modules/nodejs/manifests/install.pp:52; cannot redeclare at line 1 on node ***
```

After:
```
# puppet apply -v  -e 'class { "nodejs": npm_package_name => false,} package { "npm": ensure => "2.14.14", provider => "npm",}'
Info: Loading facts
Info: Loading facts
Notice: Compiled catalog for *** in environment production in 0.31 seconds
Info: Applying configuration version '1452278934'
Notice: /Stage[main]/Main/Package[npm]/ensure: ensure changed '1.4.29' to '2.14.14'
Notice: Finished catalog run in 4.97 seconds
```

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -103,7 +103,7 @@ class nodejs(
     validate_string($nodejs_dev_package_name)
   }
 
-  if $npm_package_name {
+  if $npm_package_name and $npm_package_name != false {
     validate_string($npm_package_name)
   }
 
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -46,7 +46,7 @@ class nodejs::install {
   }
 
   # npm
-  if $nodejs::npm_package_name {
+  if $nodejs::npm_package_name and $nodejs::npm_package_name != false {
     package { $nodejs::npm_package_name:
       ensure => $nodejs::npm_package_ensure,
     }
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -31,7 +31,7 @@ class nodejs::params {
         $nodejs_dev_package_name   = undef
         $nodejs_package_name       = 'nodejs'
         $npm_package_ensure        = 'absent'
-        $npm_package_name          = undef
+        $npm_package_name          = false
         $npm_path                  = '/usr/bin/npm'
         $repo_class                = '::nodejs::repo::nodesource'
       }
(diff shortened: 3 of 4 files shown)


igalic:
mmhm. this makes sense.
@x3dfxjunkie would you mind doing a [proper rebase](https://www.git-scm.com/book/en/v2/Git-Branching-Rebasing) (rather than that upstream merge)

****

n.b.: this would've been easier (or rather: less confusing) if you had used the [github flow](https://guides.github.com/introduction/flow/index.html) work flow.
see also https://github.com/voxpupuli/puppet-jira/pull/99#issuecomment-152528108 for a step by step how to

juniorsysadmin:
Hmmm. I'm a bit uncomfortable mixing strings and booleans in the same variable. Also, given ```undef`` evaluates to false, and therefore we don't manage the package, why are we getting duplicate declarations?

x3dfxjunkie:
At least with puppet 3.x, undef isn't the same as false. That may be more of a bug with puppet, but not sure. I could change it to a proper string (like blank or none), but that seems less intuitive.

x3dfxjunkie:
Rebased the merge out, sorry for that.

igalic:
thank you @x3dfxjunkie!

PR #200, opened 2016-01-24
Modulesync

juniorsysadmin:
Pins Rubocop so we don't have unexpected breakage


juniorsysadmin:
The fix up of actually adding 2.3.0 to the Travis testing matrix with allowed_failures has been done in https://github.com/voxpupuli/modulesync_config/pull/66 as well as here

jyaworski:
LGTM.

PR #201, opened 2016-02-04
ensure apt-get update is executed before installing packages

saz:
If you're using the nodesource repo, nodejs will be installed before apt-get update is run. This will result in the wrong version of nodejs.

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -17,12 +17,14 @@ class nodejs::install {
   # nodejs
   package { $nodejs::nodejs_package_name:
     ensure => $nodejs::nodejs_package_ensure,
+    tag    => 'nodesource_repo',
   }
 
   # nodejs-development
   if $nodejs::nodejs_dev_package_name {
     package { $nodejs::nodejs_dev_package_name:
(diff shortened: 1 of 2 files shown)


bastelfreak:
:+1:

juniorsysadmin:
Doesn't this depend on the update frequency hash? https://github.com/puppetlabs/puppetlabs-apt#class-apt

saz:
The mentioned frequency is only for configuring a run of apt-get update on each run (always), daily or weekly. This does not ensure, that apt-get update is executed, if a new source is added. This might happen by coincidence. And I don't see a reason to run an update with every puppet run. But even with the frequency set to always the module might install the wrong nodejs version, which *might* be fixed on the next run. 

Via Boxer senden

Am 04.02.2016 12:24 nachm. schrieb Ben S <notifications@github.com>:

Doesn't this depend on the update frequency hash? https://github.com/puppetlabs/puppetlabs-apt#class-apt

saz:
Is there any roadmap for a new release? I'm able to stick with the current version from github, but don't want to miss the new release to change it back to a forge version.

saz:
This PR does also resolve #171

PR #203, opened 2016-02-28
Fixes #202 - Implement a feature check for npm

### LINKED ISSUE #202: Provider should confine to feature, which should check if npm is actually installed
Currently, using provider => 'npm' without dependency on the nodejs class might result a failure.

imriz:


jyaworski:
Can you give a description of the use-case for this?

imriz:
Sure - When npm is being installed within the same catalog. This defers the use of the provider until the npm command becomes available, without requiring the user to explicitly set a dependency (i.e. require => Package['npm'] or require => Class['nodejs']).

See https://docs.puppetlabs.com/guides/provider_development.html#suitability

jyaworski:
Neat. Thanks for the contribution.

petems:
Awesome! :smile:

imriz:
Unfortunately, there is a small issue with it. 
I opened a bug report with puppetlabs on it (https://tickets.puppetlabs.com/browse/PUP-5985), but I have a workaround for which I will open another PR tomorrow.

PR #216, opened 2016-04-19
Manage /root/.npmrc

ghoneycutt:
This is necessary to specify the _auth config key which cannot be set
using 'npm config set _auth'.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -12,6 +12,7 @@ class nodejs(
   $npm_package_ensure          = $nodejs::params::npm_package_ensure,
   $npm_package_name            = $nodejs::params::npm_package_name,
   $npm_path                    = $nodejs::params::npm_path,
+  $npmrc_auth                  = $nodejs::params::npmrc_auth,
   $repo_class                  = $nodejs::params::repo_class,
   $repo_enable_src             = $nodejs::params::repo_enable_src,
   $repo_ensure                 = $nodejs::params::repo_ensure,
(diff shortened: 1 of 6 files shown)


ghoneycutt:
I'm using Artifactory with an internal NPM registry. Part of that configuration is specifying `_auth`, though this cannot be done using `npm config set _auth`, which is nice, as it would expose the credentials. This PR is meant to allow that functionality.

The README states that the module manages the npmrc, which it does not, until now, though it specifically manages /root/.npmrc.

manifests/init.pp
```diff
@@ -107,6 +108,12 @@
     validate_string($npm_package_name)
   }
 
+  if $npmrc_auth {
+    if is_string($npmrc_auth) == false {
```
ghoneycutt: Did this instead of validate_string because I wanted to check for the Puppet::Error and a specific error string which can change quite a bit using `validate_string()` on different platforms.

manifests/install.pp
```diff
@@ -55,5 +58,13 @@
       tag    => 'nodesource_repo',
     }
   }
-}
 
+  file { 'root_npmrc':
```
jyaworski: Do we want to manage the whole file? Can this be file_line?

ghoneycutt: I did this because the docs read that this file is managed, even though it actually is not. So this is a step to get reality to match the docs.

I don't have any strong feelings between managing the contents of the file as it is now or using file_line.

jyaworski: If the docs read that the whole file is managed, how do you think we should do that? I'm thinking a concat resource.

ghoneycutt: I think we should do what's in the PR :D

or file_line if you'd like.. whatever get's this merged.

ghoneycutt:
@jyaworski Thank you!

Could you please create a new tagged release with this commit.

PR #220, opened 2016-04-21
Document NPM's behavior of adding a trailing slash to URL's

ghoneycutt:
This could save a lot of headaches wondering why Puppet is not behaving
in an idempotent fashion.

### DIFF
--- a/README.md
+++ b/README.md
@@ -294,12 +294,16 @@ nodejs::npm { 'remove all express packages':
 
 nodejs::npm::global_config_entry can be used to set/remove global npm configuration settings.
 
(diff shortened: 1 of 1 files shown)


ghoneycutt:
Isn't that nice of npm :D

ghoneycutt:
@jyaworski Could you please merge this one. It bit me this morning

ghoneycutt:
Thank you!

igalic:
couldn't we do this normalisation ourselves in the code?
types have `munge`… not sure how to do that with defines though…

PR #221, opened 2016-04-21
Fix ordering bug when using global_config_entry and managing npm

ghoneycutt:


ghoneycutt:
When setting npm_package_ensure to present, without this commit, global_config_entry would run without npm being installed and error. This fixes that.

ghoneycutt:
Thank you!

PR #223, opened 2016-05-08
modulesync 0.6.0

bastelfreak:


spec/classes/nodejs_spec.rb
```diff
@@ -40,6 +40,7 @@
         {
           lsbdistcodename: lsbdistcodename,
           lsbdistid: lsbdistid,
+          lsbdistrelease: operatingsystemrelease,
```
bastelfreak: I'm not sure if this is 100% correct.

PR #230, opened 2016-06-02
Release 2.0.1

bastelfreak:


bastelfreak:
modulesync is a bigger change, needs to be done later.

PR #233, opened 2016-06-13
Update dependencies to include puppetlabs/apt

bbriggs:


juniorsysadmin:
@bbriggs puppetlabs/apt is a *soft* dependency rather than a hard one, which is why it is outlined in the README and not put in the metadata.json

See: https://github.com/voxpupuli/puppet-nodejs/blob/master/README.md#module-dependencies
and the Style Guide: https://docs.puppet.com/guides/style_guide.html#dependencies

juniorsysadmin:
It would be great if there was a way to do OS-dependent dependencies, or list soft dependencies in the metadata.json, but there is not at the moment.

PR #234, opened 2016-06-16
WIP: Validate soft dependencies exist in params.pp

nibalizer:
As a potential way of helping the user with soft dependencies, we can
attempt to load the module metadata. If we don't get any metadata back,
that probably means the module is not installed. We can then warn the
user that they need to install it.

We can also inspect returned metadata for version and author
compatibility, though ideally we would be very very lax.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -26,6 +26,13 @@ class nodejs::params {
       if $::operatingsystemrelease =~ /^6\.(\d+)/ {
         fail("The ${module_name} module is not supported on Debian Squeeze.")
       }
+      # validate apt module exists
+      $apt_data = load_module_metadata('apt', true)
+      if $apt_data = {} {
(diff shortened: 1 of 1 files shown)


juniorsysadmin:
Just a few thoughts:
1) This requires a very recent version of stdlib
2) If you're going to do this you should do warn (ideally do notice) rather than fail, otherwise the code behaviour has the same effect as making them a hard dependency
3) You could go further and use a fact (leave it up to the user to set it), ```$::check_dependencies``` true/false
4) This is still a hack around a clear direction for soft dependencies. If both hard and soft dependencies are thrown in metadata.json it's still going to cause potential issues for things like librarian-puppet and r10k (if r10k gets an option to do similar behaviour)

igalic:
if this is gonna be a very recent version of stdlib, we could just as well add a function wrapper that

Issue #238, opened 2016-07-10
incorrect repo validation for Ubuntu 15.04/15.10


drkp:
It looks like b8c672d0 changed the validate_re's for 15.04 and 15.10 to have a trailing x, so they'll only match 4.xx not 4.x. This doesn't seem right...

bastelfreak:
Hi @drkp,

thanks for reporting this. Are you able to also provide a patch for it?

drkp:
Sure, just submitted #239

juniorsysadmin:
Fixed by #239

PR #239, opened 2016-07-10
Fix repo validation regexps for Ubuntu 15.04/15.10 (#238)

drkp:


bastelfreak:
thanks for the PR. Can you add tests so we ensure that the regex is now working correctly?

jyaworski:
Ping @drkp

juniorsysadmin:
Validated from other sources, merging this one without additional tests 🔥

PR #240, opened 2016-07-10
modulesync & cleanup

bastelfreak:


igalic:
sudden realization: the only way we can "release" modulesync_config 1.x is when rubocop becomes "stable" — or else every release would be a major version bump

bastelfreak:
There weren't that many new things in the last rubocop releases. Our config was just very lax for a very long time. Fixing cops in the future will be way easier when our current config is already applied to a module.

---

Anybody has suggestions for the travis failures?

spec/unit/puppet/provider/package/npm_spec.rb
```diff
@@ -2,67 +2,69 @@
...
+      provider.expects(:npm).with('install', '--global', 'express')
+      provider.install
     end
 
     describe 'and a source is specified' do
-      it 'should use the source instead of the gem name' do
-        @resource[:source] = '/tmp/express.tar.gz'
-        @provider.expects(:npm).with('install', '--global', '/tmp/express.tar.gz')
-        @provider.install
+      it 'uses the source instead of the gem name' do
```
bbriggs: NPM packages aren't gems

bbriggs:
Best explanation I've seen so far on stubs/doubles: https://relishapp.com/rspec/rspec-mocks/v/3-2/docs/verifying-doubles/using-an-instance-double

bbriggs:
Whatever object is being tested in [57,58,65] needs to be initialized as an instance double, then tested using more normal spec language. Example:

```ruby
require 'unit_helper'

require 'user'

RSpec.describe User, '#suspend!' do
  it 'notifies the console' do
    notifier = instance_double("ConsoleNotifier")

    expect(notifier).to receive(:notify).with("suspended as")

    user = User.new(notifier)
    user.suspend!
  end
end
```

bastelfreak:
I still have no idea how we should handle these rubocop errors. Can anybody provide some guidance?

igalic:
my reccomendation is to open an issue directly in https://github.com/bbatsov/ruby-style-guide/issue

petems:
@bastelfreak I can do some fixing, had a similar rubocop stubbing issue with another module, tag me to remind me if I haven't opened a PR in a few days 👍

petems:
Ok, I looked at the thing, turns out I'm not sure how to fix it...

Should we just put a control comment for the stubs and merge for now?

bastelfreak:
+1000. I will do that after lunch

igalic:
thanks everyone!

PR #241, opened 2016-08-08
fixed to work with ruby 1.8 with old hash arrow style

knackjax:
fixed to work with ruby 1.8 with old hash arrow style


knackjax:
I cleaned up the commit messages.

bastelfreak:
Hi,
ruby187 is really outdated. We decided to drop it a few months ago. If you still rely on it please use an older release of our modules. You should really consider upgrading your environment. Every modern distribution ships a recent ruby version, also the latest puppet AIO releases have a bundled ruby and openssl package. Feel free to join our IRC channel #voxpupuli on freenode if you've further questions. We're happy to assist you during an upgrade to a newer distribution or Puppet4.

PR #244, opened 2016-08-30
Ignore npm cache lines when calling 'npm view' for latest version

domcleal:
When the npm cache has expired or is empty, npm (version 1.3.6 at least)
outputs requests it's making to stderr which are interpreted as a
version number from the `npm view` command.

If using ensure=latest, this causes the package to be reinstalled every
time the cache expires:

    Debug: Executing: '/bin/npm view phantomjs version'
    Debug: /Package[phantomjs]/ensure: phantomjs "2.1.7" is installed, latest is "npm http GET https://registry.npmjs.org/phantomjs\nnpm http 200 https://registry.npmjs.org/phantomjs\n2.1.7"
    Debug: Executing: '/bin/npm install --global phantomjs'
    Debug: /Package[phantomjs]/ensure: phantomjs "2.1.7" is installed, latest is "npm http GET https://registry.npmjs.org/phantomjs\nnpm http 200 https://registry.npmjs.org/phantomjs\n2.1.7"
    Notice: /Package[phantomjs]/ensure: ensure changed '2.1.7' to 'npm http GET https://registry.npmjs.org/phantomjs
    npm http 200 https://registry.npmjs.org/phantomjs
    2.1.7'

The npm logging lines are now filtered out similarly to the `npm list`
command.

### DIFF
--- a/lib/puppet/provider/package/npm.rb
+++ b/lib/puppet/provider/package/npm.rb
@@ -53,7 +53,7 @@ Puppet::Type.type(:package).provide :npm, parent: Puppet::Provider::Package do
   end
 
   def latest
-    npm('view', resource[:name], 'version').strip
+    npm('view', resource[:name], 'version').lines.reject { |l| l.start_with?('npm') }.join("\n").strip
   end
 
   def update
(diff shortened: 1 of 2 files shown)


juniorsysadmin:
:+1: 
For non-global npm installs under npm.pp we use execs and I believe always re-install if ensure is set to latest. Not sure if we can do a similar thing there.

PR #247, opened 2016-09-28
Set legacy_debian_symlinks true for Ubuntu 16.04

frozenfoxx:
On Ubuntu 16.04 the system will no longer update the /etc/alternatives/node by default.  This results in applications like Yeoman not being able to execute with a `sh: 1: node: not found` error.  Setting the value to `true` solves the problem.

This problem is also described here:  [http://stackoverflow.com/questions/21168141/cannot-install-packages-using-node-package-manager-in-ubuntu](http://stackoverflow.com/questions/21168141/cannot-install-packages-using-node-package-manager-in-ubuntu).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,8 @@ Gemfile.lock
 Gemfile.local
 vendor/
 .vendor/
-spec/fixtures/
+spec/fixtures/manifests/
+spec/fixtures/modules/
 .vagrant/
 .bundle/
 coverage/
--- a/.msync.yml
+++ b/.msync.yml
@@ -1,1 +1,1 @@
-modulesync_config_version: '0.11.1'
+modulesync_config_version: '0.12.8'
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -4,481 +4,482 @@ AllCops:
   Include:
     - ./**/*.rb
   Exclude:
+    - files/**/*
     - vendor/**/*
     - .vendor/**/*
     - pkg/**/*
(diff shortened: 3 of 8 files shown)


juniorsysadmin:
@frozenfoxx Just a few things:
1) Can you (or anyone else) confirm that in previous Ubuntu versions, the symlink was created for node to nodejs by default (with the nodejs legacy or other means) ? A change to default behaviour is a breaking change.
2) Are you sure that the npm removal is still happening with latest git master? (Is it not fixed by #242 ?) If so, this should be in a separate PR
3) The way you've done your change at the moment will result in duplicate declarations, no?
4) TravisCI is failing due to a modulesync needing to happen. Once that is done, and you have rebased, then the results should be a direct consequence of your changes.

frozenfoxx:
1) I cannot confirm as I currently only have production systems running old Ubuntu versions.  From reading through other guides on setting up NodeJS on Ubuntu 12.04 and 14.04 it seems that the symlink still had to be made manually.

2) Positive npm removal is still happening.  If I set the aforementioned value to present via hieradata then it ensures npm is properly installed across multiple runs.  If it is left to the default then it ensures removal.

3) Can we not redefine variable values based on conditionals later?  The only other ways I can think of offhand that's recommended is to set up a case individually for each variable or move the variable definition into every case.  I don't mind doing that if desired but it seemed like defining the default above was more desirable.

4) I'm sorry for my ignorance but I'm not familiar with a modulesync or what that means.  Where can I read more about what needs to be done?

PR #248, opened 2016-09-30
Fix for NPM and Node.JS installation on Ubuntu 16.04

frozenfoxx:
On Ubuntu 16.04 the system will no longer update the /etc/alternatives/node by default.  This results in applications like Yeoman not being able to execute with a `sh: 1: node: not found` error.  Setting the value to `true` solves the problem.  This is a revised pull request (in place of [https://github.com/voxpupuli/puppet-nodejs/pull/247}(https://github.com/voxpupuli/puppet-nodejs/pull/247) with new modulesync attempt, explicit definition of the legacy debian symlink and nodejs dev package variables.

This problem is also described here:  [http://stackoverflow.com/questions/21168141/cannot-install-packages-using-node-package-manager-in-ubuntu](http://stackoverflow.com/questions/21168141/cannot-install-packages-using-node-package-manager-in-ubuntu).  I cannot confirm its behavior on older Ubuntu distributions but reliably on 16.04 npm is actively removed and node's symlink not created, even when using the master from the repo.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,8 @@ Gemfile.lock
 Gemfile.local
 vendor/
 .vendor/
-spec/fixtures/
+spec/fixtures/manifests/
+spec/fixtures/modules/
 .vagrant/
 .bundle/
 coverage/
--- a/.msync.yml
+++ b/.msync.yml
@@ -1,1 +1,1 @@
-modulesync_config_version: '0.11.1'
+modulesync_config_version: '0.12.9'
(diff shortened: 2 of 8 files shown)


bastelfreak:
Hi @frozenfoxx,
thanks for the PR. Can you please rebase against our current master and drop the modulesync commit?

bastelfreak:
Hi, we included this commit in #251

frozenfoxx:
Thanks very much for taking care of this, I was having some real trouble figuring out how to get it to work with the modulesync.

bastelfreak:
sure sure. When you've issues again just ping us in #voxpupuli on freenode. We normally try to take care of modulesync, so contributors don't have to deal with it.

PR #255, opened 2016-10-25
Metric/BlockLength -> Metrics/BlockLength

bastelfreak:


bbriggs:
What this do?

bastelfreak:
We've disabled the cop since a longer time, but we had a typo. This is already fixed in modulesync_config, but I didn't want to make a release of it now.

bbriggs:
Hah, yeah, I saw that PR just after I asked about this one.

PR #259, opened 2016-11-16
Use http instead of https (sic) for apt source

mpdude:
This is a workaround for nodesource/distributions#388. `deb.nodesource.com` moved to CloudFront and some Linux distros (like Ubuntu 14.04) are unable to fetch packages from there because SSL with SNI is required.

A work-around is to fall back to HTTP for the time being. Note that this supposed not to impair package integrity checking, although I cannot tell whether this makes sense if the key is retrieved over HTTP as well.


mpdude:
@ghusse indeed that seems to work!

mpdude:
Looks like a fix for nodesource/distributions#388 is already being rolled out, might obsolete this PR.

juniorsysadmin:
Underlying issue fixed by Nodesource

PR #260, opened 2016-11-17
provider: add support for install_options

cbaoxa:
Use case:
pass "--loglevel=error" to npm, in order to hide warnings for deprecated dependencies during CI runs

(eg. gulp 3, https://github.com/gulpjs/gulp/issues/1806)


bastelfreak:
thanks!

PR #271, opened 2017-01-08
Don't validate repo_url_suffix

### LINKED ISSUE #270: Provide NodeJS 7.x
Could you please update the project to allow installation of Node 7.0 and higher? When I set `repo_url_suffix` to `7.x` I get the following error with puppet-nodejs version 2.2.0 (using Debian Jessie):

```
Error: Var $repo_url_suffix with value '_7.x' is not set correctly for Debian 8.6. See README. at /tmp/vagrant-puppet/modules-3da61f48bd739b819fd7ba4c41983cc2/nodejs/manifests/init.pp:52
```

Thanks!

juniorsysadmin:
repo_url_suffix is just a reflection of available NodeSource releases,
which could change at any time. Remove regex validation of this
parameter and also add changes to the README to make this clear to the
user.

Also fixes #270


juniorsysadmin:
I know that EOL operating systems have now been added to the README, but leaving them in there anyway.

README.md
```diff
@@ -433,29 +433,36 @@ User for the proxy used by the repository, if required.
...
-    * 4.x (**Only** available for RedHat/CentOS/Amazon Linux 7 and Fedora 21/22/23)
-    * 5.x (**Only** available for RedHat/CentOS/Amazon Linux 7 and Fedora 21/22/23)
+Defaults to ```0.10``` which means that the latest NodeSource 0.10.x release
+is installed. If you wish to install a 0.12.x release or greater, you will
+need to set this value accordingly. This parameter is a just a reflection of
+the NodeSource URL structure - NodeSource might remove old versions (such as
+0.10 and 0.12) or add new ones (such as 8.x) at any time.
+
+The following are ``repo_url_suffix`` values that reflect NodeSource versions
+that were available on 2016-01-08:
```
rnelson0: I think you mean 2017 :)

juniorsysadmin: Oh dear

PR #282, opened 2017-03-02
DO NOT MERGE: Prevent provider blowing up on ruby 1.8 agents

alexjfisher:
To be clear, this is not about restoring ruby 1.8 support.  This module
is now puppet 4 only and will not be able to be used on ruby 1.8
systems.

However, some puppet environments contain a mix of puppet 4 (supported)
and older puppet 3 EL6 agents. We don't want unrelated puppet runs to
fail on those old systems just because this module is in the same
environment.

CURRENTLY UNTESTED. DO NOT MERGE

### DIFF
--- a/lib/puppet/provider/package/npm.rb
+++ b/lib/puppet/provider/package/npm.rb
@@ -1,28 +1,30 @@
+# We don't support using this module on puppet 3/ruby 1.8 systems.
+# However, some puppet environments contain a mix of puppet 4 (supported) and older puppet 3 EL6 agents.
+# We don't want puppet runs to fail on those old systems just because this module is in the same environment.
+# rubocop:disable Style/HashSyntax
 require 'puppet/provider/package'
 
-Puppet::Type.type(:package).provide :npm, parent: Puppet::Provider::Package do
+Puppet::Type.type(:package).provide :npm, :parent => Puppet::Provider::Package do
   desc 'npm is the package manager for Node.js. This provider only handles global packages.'
 
-  confine feature: :npm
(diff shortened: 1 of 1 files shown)


anordby:
Hi. Fails here, with Puppet 3.8.7-1 in RHEL 6 with Ruby 1.8.7. Error log attached.
[npm.rb.errors.txt](https://github.com/voxpupuli/puppet-nodejs/files/815191/npm.rb.errors.txt)

alexjfisher:
@anordby Sorry, that was very naive of me to think that confining the provider to ruby > 1.8 would mean I wouldn't have to change the hash syntax in all of the methods.  Let me try again...

roidelapluie:
@bastelfreak ? If we do this do we need this on all modules?

alexjfisher:
@roidelapluie This particular module might be a bit special as it adds a provider to the core `package` type?  The user didn't want or expect to be able to *use* it with their old server, but did want to be able to have it in the environment (for use by newer servers).

We've made similar concessions in the past.  eg https://github.com/voxpupuli/puppet-iis/pull/111

roidelapluie:
OK let's do this

roidelapluie:
Did not remember that this was «DO NOT MERGE»

alexjfisher:
@roidelapluie It was only still 'DO NO MERGE' in the sense I knew it might be contentious.  Everything should otherwise be fine.  I'll rename the PR for the sake of the next changelog.

roidelapluie:
Ok.

Issue #281, opened 2017-02-23
Dependency cycle on ubuntu 14.04


pscadiz:
## Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: 3.8
- Ruby: 2.0.0
- Distribution: Ubuntu 14.04 (Trusty)
- Module version: (used latest commit)

## How to reproduce (e.g Puppet code you use)

```puppet
class { '::nodejs':
}
```

## What are you seeing
Puppet fails to run with a dependency cycle

## What behaviour did you expect instead
Node to be installed

## Output log
```
Error: Could not apply complete catalog: Found 1 dependency cycle
(Anchor[::nodejs::begin] => Class[Nodejs::Repo::Nodesource] => Class[Nodejs::Repo::Nodesource::Apt] => Apt::Source[nodesource] => Apt::Setting[list-nodesource] => File[/etc/apt/sources.list.d/nodesource.list] => Class[Apt::Update] => Exec[apt_update] => Class[Apt::Update] => Stage[setup] => Stage[main] => Class[Nodejs::Repo::Nodesource])
```

## Any additional information you'd like to impart
Running puppet on a docker container based on atlassian/bamboo-base-agent. Also updated puppetlabs-apt to tag 2.3.0

zehzehpeh:
:+1:
Confirmed on : 
- Puppet: 3.8.6
- Ruby: 2.3.1
- Distribution: Ubuntu 16.04 (Xenial)
- Module version: 2.3.0
- `puppetlabs-apt` at 2.3.0
- not running on a docker container
- also happens when using `repo_url_suffix => '6.x',`

zehzehpeh:
hey! Any update or ETA on this?

bastelfreak:
Hi @pscadiz and @zehzehpeh. Thanks for reporting this error. Is one of you able to provide a patch for this?

zehzehpeh:
@bastelfreak unfortunately I am not able to invest much time in this at the moment. However, I'd be more than happy to test any patch you send my way.

ffrank:
I'd love to see this fixed, but please note that Puppet 3.x is not supported by this module anymore. The issue appears to not be reproducible w/ Puppet 4.x.

Is a Puppet upgrade an option for you at all?

zehzehpeh:
>  Puppet 3.x is not supported by this module anymore.

Fair enough.

> Is a Puppet upgrade an option for you at all?

[...]

PR #286, opened 2017-03-23
Fix dependency cycle error

zehzehpeh:
This fixes the dependency cycle error reported in issue #281 .
Since `ensure_packages` is used, the `require` statement is obsolete.

Tested on Ubuntu 16.04.2 LTS.

### DIFF
--- a/manifests/repo/nodesource/apt.pp
+++ b/manifests/repo/nodesource/apt.pp
@@ -23,10 +23,6 @@ class nodejs::repo::nodesource::apt {
       pin      => $pin,
       release  => $::lsbdistcodename,
       repos    => 'main',
-      require  => [
-        Package['apt-transport-https'],
-        Package['ca-certificates'],
-      ],
     }
 
     Apt::Source['nodesource'] -> Package<| tag == 'nodesource_repo' |>


juniorsysadmin:
@zehzehpeh After talking to a number of other people on this, I'm pretty sure that that ensure_packages does not affect dependency ordering. Are you certain that this was the only change that fixed your issue?

zehzehpeh:
I tested it anew, and yes : without this change, the puppet run fails right after catalog compilation with the following error : 
```
Error: Failed to apply catalog: Found 1 dependency cycle:
(Anchor[apt_key 9FD3B784BC1C6FC31A8A0A1C1655A0AB68576280 present] => Apt::Key[Add key: 9FD3B784BC1C6FC31A8A0A1C1655A0AB68576280 from Apt::Source nodesource] => Apt::Setting[list-nodesource] => File[/etc/apt/sources.list.d/nodesource.list] => Class[Apt::Update] => Exec[apt_update] => Class[Apt::Update] => Package[apt-transport-https] => Apt::Source[nodesource] => Apt::Setting[list-nodesource])
```
Yet with __only__ this change, the runs completes successfully.

ffrank:
I don't think I agree with the notion that `ensure_packages` eliminates the need to express proper relationships. Could you elaborate on that?

The fact that people have not been able to pinpoint the reason for the cycle, in connection with the observation that I cannot reproduce the original bug with Puppet 4.2, I'm inclined to assume that this is an issue with the implementation of `contain` in older versions of Puppet.

zehzehpeh:
> I don't think I agree with the notion that `ensure_packages` eliminates the need to express proper relationships. Could you elaborate on that?

`ensure_packages` will install the packages if not present. And `require => Package` will do the same thing. The ordering will also stay the same, since `ensure_packages` is declared before `apt::source {}`.

OTOH, it is true that `require => Package` makes it very clear that those packages are needed for `apt::source {}`, so it is clearer to read. But not technically necessary.

Or am I missing something?

ffrank:
Actually, yes. The `require => Package[name]` establishes a resource relationships. The agent is guaranteed to apply the required resource first, and if the required resource fails, the requiring resource is skipped altogether.

The lexical order in the manifest is largely ignored by Puppet before 4.0 by default, and the order of resource application is pseudo-random. In recent versions, Puppet *will* order resources the way they appear in the manifest by default. This will not establish the fail/skip semantics outlined above, but this is a secondary issue here.

The point is: Dropping a relationship might mix up the order in some manifests that use the module, and Puppet 3.x, even if it does not do that in your tests. So I'm wary.

If you cannot resolve this one through a Puppet update, I suggest the following: Add a parameter to the module that allows you to selectively disable this relationship. It would be implemented kind of like this:

```
unless $skip_strict_requires {
  Package['x', 'y'] -> Apt::Source['z']
}
```

(Although I hope we can avoid adding this code, as it would be a hack for Puppet versions that we do not really want to actively support anymore.)

zehzehpeh:
> The point is: Dropping a relationship might mix up the order in some manifests that use the module, and Puppet 3.x, even if it does not do that in your tests. So I'm wary.

Fair point. Thanks for the details!

> a hack for Puppet versions that we do not really want to actively support anymore

I am not going to PR that. After testing on my side, it seems the problem is indeed related to Puppet 3 vs 4.

Issue #290, opened 2017-04-11
NPM Proxy


Poil:
Hi,

We used internal repository for all OS packages
But for NPM packages we don't mirror it, and we need to set a proxy.
Today it does not seem possible to have no-proxy on "repo" but a proxy on NPM.
We also manage "/root/.npmrc" but it only take _auth into account (no way to add other parameter), perhaps it will be easier to allow a hash that contains keys=>values

Best regards,

juniorsysadmin:
Fixed by #292

PR #291, opened 2017-04-11
npmrc can take k,v

Poil:
Fix for #290

Add a npmrc_config hash to allow the possibility to add some configuration to /root/.npmrc

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -13,6 +13,7 @@ class nodejs(
   $npm_package_name            = $nodejs::params::npm_package_name,
   $npm_path                    = $nodejs::params::npm_path,
   $npmrc_auth                  = $nodejs::params::npmrc_auth,
+  $npmrc_config                = $nodejs::params::npmrc_config,
   $repo_class                  = $nodejs::params::repo_class,
   $repo_enable_src             = $nodejs::params::repo_enable_src,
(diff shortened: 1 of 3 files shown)


bastelfreak:
Hi @Poil, thanks for your pull request!
You used the email address benjamin.dupuis@fr.clara.net, which isn't associated with your github account. Can you please check if this address is valid and add it to your github profile?

manifests/init.pp
```diff
@@ -13,6 +13,7 @@
   $npm_package_name            = $nodejs::params::npm_package_name,
   $npm_path                    = $nodejs::params::npm_path,
   $npmrc_auth                  = $nodejs::params::npmrc_auth,
+  $npmrc_config                = $nodejs::params::npmrc_config,
```
bastelfreak: please use a puppet4 datatype here. Probably work will:

```puppet
Optional[Hash] $npmrc_config                = $nodejs::params::npmrc_config,
```

manifests/init.pp
```diff
@@ -49,6 +50,11 @@
       fail('npmrc_auth must be a string')
     }
   }
+  if $npmrc_config {
```
bastelfreak: the whole block isn't needed if you use the datatype

PR #293, opened 2017-04-11
Fix puppetlint

Poil:
just a little fix for puppetlint cf. https://docs.puppet.com/puppet/latest/style_guide.html#chaining-arrow-syntax


bastelfreak:
thanks!

PR #292, opened 2017-04-11
npmrc can take k,v

Poil:
Fix for #290

Add a npmrc_config hash to allow the possibility to add some configuration to /root/.npmrc

PR #294, opened 2017-04-11
Npmrc config

Poil:
Fix for #290

Add a npmrc_config hash to allow the possibility to add some configuration to /root/.npmrc

I tried to add a test (never did this before so perhaps it's not the right way)


juniorsysadmin:
#292

PR #298, opened 2017-04-21
Add the ability to manage dependent packages else where.

kwolf:
We already manage one of the packages (`ca-certificates`) elsewhere.  This adds the ability to manage dependent packages outside of the nodejs module to avoid duplicate package definitions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -469,6 +469,12 @@ that were available on 2017-01-08:
 The USE flags to use for the Node.js package on Gentoo systems. Defaults to
 ['npm', 'snapshot'].
 
+#### `manage_dep_packages`
+
+Specify if hte puppet-nodejs module should install dependent packages
+required by nodejs. If you manage these packages elsehwere, set to
+false.  Defaults to `true`.
+
 ## Limitations
 
 This module has received limited testing on:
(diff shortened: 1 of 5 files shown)


juniorsysadmin:
@kwolf Would prefer that this uses `manage_dependencies` as the parameter for semi-consistent variable naming reasons

kwolf:
I thought about that but I wanted it to be more clear that it was package dependencies.  I am still happy to change it if you want me too.

juniorsysadmin:
Puppet will pull in package dependencies via yum, apt etc so leaving package in the name might lead to confusion. ```manage_repo_dependencies``` is even more explicit, if you were after more naming suggestions...

README.md
```diff
@@ -469,6 +469,12 @@ that were available on 2017-01-08:
 The USE flags to use for the Node.js package on Gentoo systems. Defaults to
 ['npm', 'snapshot'].
 
+#### `manage_dep_packages`
```
juniorsysadmin: Make it clear in the README that this refers to requirements for an apt HTTPS repository, not nodejs itself.

juniorsysadmin:
@kwolf Status of this PR?

juniorsysadmin:
Kinda superseded by #348 . Which might make it more difficult to solve your problem. Or not.

Issue #167, opened 2015-09-21
install dependencies from package.json


rougeth:
Is there a way to use this module  to install dependencies on a `package.json` file?

igalic:
```puppet
exec { 'install deps for my wintersmith blog'
  command => 'npm install',
  cwd     => '/srv/web/blog',
  creates => 'node_modules',
  path    => [ '/bin', '/usr/bin', '/usr/local/bin' ],
}
```

rougeth:
@igalic thanks

juniorsysadmin:
Also see #154 for a short discussion on this.

Issue #154, opened 2015-07-26
Running `npm install` in a package dir with no arguments


ingomueller-net:
I would like to run `npm install` in a package dir with no arguments in order to install the dependencies of this package. I do not want to deploy the package dir itself with npm. This corresponds to the first form of invocation of npm in [the docs](https://docs.npmjs.com/cli/install). I believe that this is currently not possible, right? I suggest to add this functionality though...

juniorsysadmin:
An empty string as the title for a defined type is not valid, so no, you can't do this with the npm defined type at the moment. This should probably be a new type of its own. Say, `run_npm_install { '/path/to/directory': }`

igalic:
+1

serviceman:
+1

igalic:
@serviceman do you wanna provide a patch?
me and @juniorsysadmin are mostly covered in administration these days :(

serviceman:
wish so, but unfortunately I have few time and zero knowledge of Ruby (and rather few of puppet), so I can't make it!

jyaworski:
What's the status of this?

igalic:
@jyaworski this is still desired and missing

NoodlesNZ:
While it's not ideal, you can run nodejs::npm like this:

```
nodejs::npm { 'App dependencies':
  ensure  => 'present',
  package => undef,
  target  => '/home/example/',
}
```
Which runs "npm install" in that directory

igalic:
is that idempotent, though?

juniorsysadmin:
Functionality added by #300

PR #300, opened 2017-05-19
Adding an option to install npm deps from package.json

poikilotherm:
Dear maintainers of voxpopuli,

I found #154 and #167 and needed the functionality for myself (installing a Mathoid server, which is based on NodeJS).

Thus I added the necessary bits and would be glad if this would be merged upstream. From my point of view, the code should be idempotent.

Thanks in advance!

Cheers,
Oliver

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,6 +149,9 @@ except version ranges. The title simply must be a unique, arbitrary value.
   specified as an array.
 * The user parameter is provided should you wish to run npm install or npm rm
   as a specific user.
+* If you want to use a package.json supplied by a module to install dependencies
+  (e.g. if you have a NodeJS server app), set the parameter use_package_json to true.
+  The package name is then only used for the resource name. source parameter is ignored.
 
 nodejs::npm parameters:
 
@@ -159,6 +162,7 @@ nodejs::npm parameters:
 * uninstall_options: option flags invoked during removal (optional).
 * npm_path: defaults to the value listed in `nodejs::params`
 * user: defaults to undef
+* use_package_json: read and install modules listed in package.json in target dir and install those in subdirectory node_modules (defaults to false)
 
 Examples:
 
@@ -286,6 +290,16 @@ nodejs::npm { 'express with options':
 }
 ```
 
+Install dependencies from package.json:
+
+```puppet
+nodejs::npm { 'serverapp':
+  ensure           => 'present',
+  target           => '/opt/serverapp',
+  use_package_json => true,
+}
+```
+
 Uninstall any versions of express in /opt/packages regardless of source:
 
 ```puppet
@@ -293,6 +307,16 @@ nodejs::npm { 'remove all express packages':
   ensure  => 'absent',
   package => 'express',
   target  => '/opt/packages',
+}
+```
+
+Uninstall dependencies from package.json:
+
+```puppet
+nodejs::npm { 'serverapp':
+  ensure           => 'absent',
+  target           => '/opt/serverapp',
+  use_package_json => true,
 }
 ```
 
--- a/manifests/npm.pp
+++ b/manifests/npm.pp
@@ -10,6 +10,7 @@ define nodejs::npm (
   Array $uninstall_options  = [],
   $home_dir                 = '/root',
   $user                     = undef,
+  Boolean $use_package_json = false,
 ) {
 
   $install_options_string = join($install_options, ' ')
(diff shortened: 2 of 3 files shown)


manifests/npm.pp
```diff
@@ -10,13 +10,15 @@
...
   $user              = undef,
+  $use_package_json  = false,
 ) {
 
   validate_re($ensure, '^[^<>=]', "The ${module_name}::npm defined type does not accept version ranges")
   validate_array($install_options)
   validate_string($package)
   validate_absolute_path($target)
   validate_array($uninstall_options)
+  validate_bool($use_package_json)
```
wyardley: If you rebase, this should now use data types instead (remove `validate_bool`, and update
`$use_package_json  = false,`
to 
`Boolean $use_package_json  = false,`
e.g., https://github.com/voxpupuli/puppet-nodejs/blob/master/manifests/init.pp#L4

manifests/npm.pp
```diff
@@ -43,21 +45,34 @@
...
-  $install_check = $::osfamily ? {
-    'Windows' => "${npm_path} ls --long --parseable | ${grep_command} \"${target}\\node_modules\\${install_check_package_string}\"",
-    default   => "${npm_path} ls --long --parseable | ${grep_command} \"${target}/node_modules/${install_check_package_string}\"",
+  $dirsep = $::osfamily ? {
+    'Windows' => "\\",
+    default   => '/'
   }
 
+  $list_command = "${npm_path} ls --long --parseable"
+  $install_check = "${list_command} | ${grep_command} \"${target}${dirsep}node_modules${dirsep}${install_check_package_string}\""
```
juniorsysadmin: I have mixed feelings about a using a variable for a directory separator in a string - I personally prefer explicit versus implicit values for it.

manifests/npm.pp
juniorsysadmin: True, but probably better to change all of these in a new PR, now this one

poikilotherm:
Hi @all,

I just rebased... Do you prefer a single commit for all changes? (I personally prefer to seperate code, tests and doc changes in different commits.)

Cheers,
Oliver

manifests/npm.pp
poikilotherm: IMHO this makes the code more readable - the string just differ by the dir seps...

Maybe let's replace it with a call to `regsubst` on Windows to replace `/` with `\\`?

juniorsysadmin: > Maybe let's replace it with a call to regsubst on Windows to replace / with \\?

That's an idea. Not sure if there will be unintended side-effects to doing this though.

juniorsysadmin:
> Do you prefer a single commit for all changes

I do. But CONTRIBUTING.md just says "squash your commits down into logical components"

poikilotherm:
Squashed as requested. Please merge :-)

manifests/npm.pp
poikilotherm: Maybe this is not even necessary...

According to https://docs.puppet.com/puppet/5.2/lang_windows_file_paths.html#when-to-use-each-kind-of-slash, we can just use slashes.

I don't have a Windows Puppet at hands to prove this true or false - anything available on your side?

juniorsysadmin: @poikilotherm Er I'm not sure if I'm much help - I haven't tested Puppet on Windows with anything for...quite some some time. Looking at the docs in more detail though, it seems as though we _do_ care about what type of slashes, because we're using cmd.exe to run the grep-alternative and cmd.exe requires backslashes.

wyardley:
@poikilotherm @juniorsysadmin I think a single commit is fine. The logical components bit is mostly just saying that it's Ok to break up the commits if they're done in a logical way.
What vox mostly tries to avoid is

* Update type to foo
* Fixed typo
* ASDF
* I don't know what this is for
* Updated spec tests

So, squashing to a single commit is usually appropriate for smaller changes. If it's a really big changeset, I have found that breaking it up into chunks can make it easier for the reviewer. However, it can make rebasing a pain sometimes too.

juniorsysadmin:
@poikilotherm Merge, or don't merge just yet?

poikilotherm:
I'm ok with the `/` and `\` thing as it is right now. If your can live with it, please go ahead and merge... :-)

manifests/npm.pp
```diff
@@ -10,6 +10,7 @@
   Array $uninstall_options  = [],
   $home_dir                 = '/root',
   $user                     = undef,
+  Boolean $use_package_json = false,
```
bastelfreak: datatypes \o/

README.md
```diff
@@ -296,6 +310,16 @@ nodejs::npm { 'remove all express packages':
 }
 ```
 
+Uninstall dependencies from package.json:
+
+```puppet
+nodejs::npm { 'serverapp':
```
bastelfreak: docs \o/

bastelfreak:
Thanks for the PR @poikilotherm!

manifests/npm.pp
igalic: `/` should just work, both in puppet and in node.

PR #304, opened 2017-06-21
Add support for Debian Stretch

rosenbergj:
Per [Nodesource](https://github.com/nodesource/distributions/blob/master/README.md), they support stretch via an alias to jessie.


rosenbergj:
Closing; nodesource added proper Stretch sources.

Issue #307, opened 2017-09-06
EPEL now has dependency on http-parser


wyardley:
Getting this error (on CentOS 7, with Puppet agent 5.x) when the package is coming from epel, including with the latest version of the module.

```
[root@c7node0 ~]# cat /etc/redhat-release 
CentOS Linux release 7.1.1503 (Core) 
[root@c7node0 ~]# puppet agent --version
5.1.0
```
(I've tried it with a more up-to-date CentOS version)
```
Error: Execution of '/bin/yum -d 0 -e 0 -y install nodejs' returned 1: Error: Package: 1:nodejs-6.11.1-1.el7.x86_64 (epel)
           Requires: http-parser >= 2.7.0
Error: Package: 1:nodejs-6.11.1-1.el7.x86_64 (epel)
           Requires: libhttp_parser.so.2()(64bit)
 You could try using --skip-broken to work around the problem
 You could try running: rpm -Va --nofiles --nodigest
Error: /Stage[main]/Nodejs::Install/Package[nodejs]/ensure: change from 'purged' to 'present' failed: Execution of '/bin/yum -d 0 -e 0 -y install nodejs' returned 1: Error: Package: 1:nodejs-6.11.1-1.el7.x86_64 (epel)
           Requires: http-parser >= 2.7.0
Error: Package: 1:nodejs-6.11.1-1.el7.x86_64 (epel)
           Requires: libhttp_parser.so.2()(64bit)
 You could try using --skip-broken to work around the problem
 You could try running: rpm -Va --nofiles --nodigest
Error: Execution of '/bin/yum -d 0 -e 0 -y install npm' returned 1: Error: Package: 1:nodejs-6.11.1-1.el7.x86_64 (epel)
           Requires: http-parser >= 2.7.0
Error: Package: 1:nodejs-6.11.1-1.el7.x86_64 (epel)
           Requires: libhttp_parser.so.2()(64bit)
 You could try using --skip-broken to work around the problem
 You could try running: rpm -Va --nofiles --nodigest
Error: /Stage[main]/Nodejs::Install/Package[npm]/ensure: change from 'purged' to 'present' failed: Execution of '/bin/yum -d 0 -e 0 -y install npm' returned 1: Error: Package: 1:nodejs-6.11.1-1.el7.x86_64 (epel)
           Requires: http-parser >= 2.7.0
Error: Package: 1:nodejs-6.11.1-1.el7.x86_64 (epel)
[...]

PR #308, opened 2017-09-07
Update acceptance tests, add EPEL test case for RedHat acceptance tests

wyardley:
This does a few things:

1. Updates `spec_helper_acceptance` to use `puppet_install_helper`
1. Updates `.travis.yml` / `.sync.yml` to run acceptance tests in CI (CentOS 7 and Ubuntu 14)
1. Verifies some extra stuff, including the package source. This will fail as of now, because of the problem mentioned in #307 / https://bugzilla.redhat.com/show_bug.cgi?id=1481008

There are still some issues that running these tests exposes. It seems like Ubuntu 16.04 installs the Ubuntu version of nodejs vs. the nodesource one; I'm not trying to fix that one (or test for it) here.

### DIFF
--- a/.sync.yml
+++ b/.sync.yml
@@ -1,3 +1,7 @@
 ---
 .travis.yml:
   secure: "SCQpiBB9qpZAjBRk+b9D3cSCQfYpDgHPOdOc7djfGeB5yn1UbGg7uW1zshrshb4QMLfUgvsL4LsT0CYj7ilBerghgeySF5JWuZdk05W/7Iudls4btbxdVjqtALR7p02mnk40qHTR1Tdb/j0gXW9uigU6nQU9iCP+Poa1KF6PXpk="
+  docker_sets:
+    - set: docker/ubuntu-14.04
+    - set: docker/centos-7
+    - set: docker/debian-8
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,25 @@ script:
 matrix:
   fast_finish: true
   include:
-  - rvm: 2.1.9
+  - rvm: 2.4.1
+    bundler_args: --without development
+    dist: trusty
+    env: PUPPET_INSTALL_TYPE=agent BEAKER_debug=true BEAKER_set=docker/ubuntu-14.04 CHECK=beaker
+    services: docker
+    sudo: required
+  - rvm: 2.4.1
+    bundler_args: --without development
+    dist: trusty
+    env: PUPPET_INSTALL_TYPE=agent BEAKER_debug=true BEAKER_set=docker/centos-7 CHECK=beaker
+    services: docker
+    sudo: required
+  - rvm: 2.4.1
+    bundler_args: --without development
+    dist: trusty
+    env: PUPPET_INSTALL_TYPE=agent BEAKER_debug=true BEAKER_set=docker/debian-8 CHECK=beaker
+    services: docker
+    sudo: required
+  - rvm: 2.4.1
     bundler_args: --without system_tests development
     env: PUPPET_VERSION="~> 4.0" CHECK=test
   - rvm: 2.2.7
--- a/spec/acceptance/class_spec.rb
+++ b/spec/acceptance/class_spec.rb
@@ -1,16 +1,47 @@
 require 'spec_helper_acceptance'
 
-describe 'nodejs class:', unless: UNSUPPORTED_PLATFORMS.include?(fact('osfamily')) do
-  it 'runs successfully' do
-    pp = <<-EOS
-    class { 'nodejs': }
-    if $::osfamily == 'RedHat' and $::operatingsystemrelease =~ /^5\.(\d+)/ {
-      class { 'epel': }
-      Class['epel'] -> Class['nodejs']
-    }
-    EOS
+describe 'nodejs class:', unless: UNSUPPORTED_PLATFORMS.include?(fact('os.family')) do
+  case fact('os.family')
+  when 'RedHat'
+    pkg_cmd = 'yum info nodejs | grep "^From repo"'
+  when 'Debian'
+    pkg_cmd = 'dpkg -s nodejs | grep ^Maintainer'
+  end
 
-    apply_manifest(pp, catch_failures: true)
-    apply_manifest(pp, catch_changes: true)
+  context 'default parameters' do
+    let(:pp) { "class { 'nodejs': }" }
+
+    it 'runs successfully' do
+      apply_manifest(pp, catch_failures: true)
+      apply_manifest(pp, catch_changes: true)
+    end
+
+    if %w[RedHat Debian].include? fact('os.family')
+      describe package('nodejs') do
+        it { is_expected.to be_installed }
+        it 'comes from the expected source' do
+          pkg_output = shell(pkg_cmd)
+          expect(pkg_output.stdout).to match 'nodesource'
+        end
+      end
+    end
+  end
+
+  context 'repo_class => epel', if: fact('os.family') == 'RedHat' do
+    let(:pp) { "class { 'nodejs': repo_class => '::epel' }" }
+
+    it 'runs successfully' do
+      apply_manifest(pp, catch_failures: true)
+      apply_manifest(pp, catch_changes: true)
+    end
+
+    describe package('nodejs') do
+      it { is_expected.to be_installed }
+      it 'comes from the expected source' do
+        pending("This won't work until we have CentOS 7.4 because of dependency")
+        pkg_output = shell(pkg_cmd)
+        expect(pkg_output.stdout).to match 'epel'
+      end
+    end
   end
 end
--- a/spec/spec_helper_acceptance.rb
+++ b/spec/spec_helper_acceptance.rb
@@ -1,21 +1,7 @@
 require 'beaker-rspec'
+require 'beaker/puppet_install_helper'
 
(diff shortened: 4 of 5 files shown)


.travis.yml
```diff
@@ -16,6 +16,18 @@ matrix:
...
+    env: PUPPET_INSTALL_TYPE=agent BEAKER_debug=true BEAKER_set=docker/ubuntu-14.04 CHECK=beaker
+    services: docker
+    sudo: required
+  - rvm: 2.1.9
+    bundler_args: --without development
+    dist: trusty
+    env: PUPPET_INSTALL_TYPE=agent BEAKER_debug=true BEAKER_set=docker/centos-7 CHECK=beaker
+    services: docker
+    sudo: required
+  - rvm: 2.1.9
```
bastelfreak: Can we use 2.4.1 here?

.sync.yml
```diff
@@ -1,3 +1,6 @@
 ---
 .travis.yml:
   secure: "SCQpiBB9qpZAjBRk+b9D3cSCQfYpDgHPOdOc7djfGeB5yn1UbGg7uW1zshrshb4QMLfUgvsL4LsT0CYj7ilBerghgeySF5JWuZdk05W/7Iudls4btbxdVjqtALR7p02mnk40qHTR1Tdb/j0gXW9uigU6nQU9iCP+Poa1KF6PXpk="
+  docker_sets:
+    - set: docker/ubuntu-14.04
```
bastelfreak: Does 16.04 also work?

wyardley: 16.04 fails if we want the source to be nodesource, as it seems like it's supposed to be:
https://github.com/voxpupuli/puppet-nodejs/blob/master/manifests/params.pp#L86
16.04 gives me:
```
Maintainer: Ubuntu Developers <ubuntu-devel-discuss@lists.ubuntu.com>
```
I'm not sure if that's a bug (probably because Ubuntu 16 version is > nodesource one), or if it's intentional. If it's intentional, we could do some logic to look for different expected package sources depending on vendor.

Debian 82 did work.

wyardley:
ps - the failed test on CentOS 7 is expected (for now).

.travis.yml
wyardley: I think this would need to come from modulesync somehow?
https://github.com/voxpupuli/modulesync_config/blob/master/moduleroot/.travis.yml.erb#L39

.sync.yml
wyardley: ps - _before_ the updates to spec_helper_acceptance, the tests were failing completely on multiple platforms, so this at least is an improvement, I think. Once we know what the expected behavior is, we can figure out how the tests should be (and if any fixes are needed in the actual module code).

bastelfreak: could you add debian 8? Does 9 work as well

wyardley: As an additional one, or instead of Ubuntu?
FWIW, I don't think modulesync provides Debian 9 yet, and I don't see it in the rake tasks
https://github.com/voxpupuli/modulesync_config/tree/master/moduleroot/spec/acceptance/nodesets

spec/acceptance/class_spec.rb
```diff
@@ -1,16 +1,61 @@
...
+  context 'default parameters' do
+    it 'runs successfully' do
+      pp = <<-EOS
+      class { 'nodejs': }
+      if $::osfamily == 'RedHat' and $::operatingsystemrelease =~ /^5\.(\d+)/ {
+        class { 'epel': }
+        Class['epel'] -> Class['nodejs']
+      }
+      EOS
+      # Run it twice and test for idempotency
```
ekohl: I like https://github.com/theforeman/foreman-installer-modulesync/blob/0ee390953528a18b48ffdf94a97b71aa5029b1ad/moduleroot/spec/spec_helper_acceptance.rb.erb#L39-L47

spec/acceptance/class_spec.rb
```diff
@@ -1,16 +1,61 @@
...
+      apply_manifest(pp, catch_failures: true)
+      expect(apply_manifest(pp, catch_failures: true).exit_code).to be_zero
+    end
+
+    if fact('os.family') == 'RedHat' || fact('os.family') == 'Debian'
```
ekohl: If you use ```['RedHat', 'Debian'].include? fact('os.family')``` then you only need to get the fact once. Since it's retrieved from the VM/container over SSH this might be a good thing to consider.

spec/acceptance/class_spec.rb
```diff
@@ -1,16 +1,61 @@
...
+        it 'comes from the expected source' do
+          pkg_output = shell(pkg_cmd)
+          expect(pkg_output.stdout).to match 'nodesource'
+        end
+      end
+    end
+  end
+
+  context 'repo_class => epel' do
+    if fact('os.family') == 'RedHat'
```
ekohl: Can you squash this into the context: ```context 'repo_class => epel', if: fact('os.family') == 'RedHat' do``` (note that I'm not 100% sure this works so please test it)

spec/acceptance/class_spec.rb
wyardley: I like this too, but can't use shared examples in an 'it' block, and maybe overkill to use within the same file? I am going to switch to the catch_failures / catch_changes pattern and simplify things a bit.

.travis.yml
wyardley: Now that it's updated in modulesync, I'll give it a shot as an extra test.

wyardley:
@ekohl @bastelfreak - How does that look now? I think I've addressed most of your (very helpful) feedback.
The tests now pass except for the expected failure.

spec/acceptance/class_spec.rb
ekohl: That does work: https://github.com/theforeman/puppet-foreman/blob/f0cfa1f39de13c1e937d23f2638db858eb7d2d27/spec/acceptance/foreman_cli_spec.rb#L33

spec/acceptance/class_spec.rb
```diff
@@ -1,16 +1,63 @@
...
+    end
+  end
+
+  context 'repo_class => epel', if: fact('os.family') == 'RedHat' do
+    let(:pp) do
+      <<-EOS
+        class { 'nodejs':
+          repo_class => '::epel',
+        }
+        class { 'epel': }
```
ekohl: I think these 2 lines are redundant (or should be): https://github.com/voxpupuli/puppet-nodejs/blob/8fc47fb1644ca1818da3cc834819286a93b7c20f/manifests/init.pp#L36-L39

wyardley: I wasn't sure about the history of that, so was hesitant to yank it otu; I will see if the test still works Ok with that removed, though, and commit if so.

We should also switch to 'contain' vs. anchor pattern (can do this in a separate PR)

ekohl: > We should also switch to 'contain' vs. anchor pattern (can do this in a separate PR)

+1

wyardley:
In most recent push, I set the failing part as pending, we could then take the pending line out when CentOS 7.4 comes out?

wyardley:
@ekohl: how's it look now?

spec/spec_helper_acceptance.rb
```diff
@@ -31,7 +17,6 @@
     # Install module and dependencies
     hosts.each do |host|
       copy_module_to(host, source: proj_root, module_name: 'nodejs')
-      shell("/bin/touch #{default['puppetpath']}/hiera.yaml")
       on host, puppet('module install puppetlabs-apt --version 2.0.1'), acceptable_exit_codes: [0, 1]
```
bastelfreak: should we remove the version foo here to get the latest?

ekohl: I'd consider https://github.com/puppetlabs/beaker-module_install_helper since that also takes are of the ```copy_module_to``` line above if you're going to fix that.

spec/acceptance/class_spec.rb
```diff
@@ -1,16 +1,47 @@
...
+  when 'RedHat'
+    pkg_cmd = 'yum info nodejs | grep "^From repo"'
+  when 'Debian'
+    pkg_cmd = 'dpkg -s nodejs | grep ^Maintainer'
+  end
+
+  context 'default parameters' do
+    let(:pp) { "class { 'nodejs': }" }
+
+    it 'runs successfully' do
```
ekohl: I would add the shared example for this but I can submit a PR for that later.

wyardley: Cool, I didn't spend too long trying to get it to work (it wouldn't work inside an 'it' block).
I did see someone else create a `self.foo` in a test, and then use that where they wanted to reuse an item in the same test (vs. across different files).

spec/spec_helper_acceptance.rb
wyardley: Fine, I improved it and now you want more improvements? 😛 Ok, sold

wyardley:
Updated to use module_install_helper

spec/acceptance/class_spec.rb
ekohl: The trick is that it already contains it blocks so you don't need that.

ekohl: https://github.com/voxpupuli/puppet-nodejs/pull/313

PR #310, opened 2017-09-07
Replace anchors with 'contain'

wyardley:
replace anchors with `contain`

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -30,13 +30,12 @@ class nodejs(
     fail("${module_name}: The manage_package_repo parameter was set to true but no repo_class was provided.")
   }
 
-  include '::nodejs::install'
+  contain '::nodejs::install'
 
   if $manage_package_repo {
     include $repo_class
-    anchor { '::nodejs::begin': }
-    -> Class[$repo_class]
+
+    Class[$repo_class]
     -> Class['::nodejs::install']
-    -> anchor { '::nodejs::end': }
   }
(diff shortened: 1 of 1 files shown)


manifests/init.pp
```diff
@@ -33,10 +33,9 @@
   include '::nodejs::install'
 
   if $manage_package_repo {
-    include $repo_class
-    anchor { '::nodejs::begin': }
-    -> Class[$repo_class]
+    contain $repo_class
```
ekohl: Including it is better. Say a user has ```Class['::epel'] -> Class['::x'] -> Class['::nodejs']``` then trying to ```contain ::epel`` here will lead to issues while it shouldn't be. If the installation is contained and you set the repo class as a dependency then you're set. We don't care where epel is (as long as it's before us) and we don't give a guarantee about it.

manifests/init.pp
```diff
@@ -33,10 +33,9 @@
   include '::nodejs::install'
```
ekohl: I would contain this.

manifests/init.pp
wyardley: What was the anchor class around it for before then? Is it not needed, even if we switch back to `include`?

wyardley: See if the version I just pushed looks better

ekohl: I think the previous implementation was incorrect and this is the correct one.

PR #311, opened 2017-09-07
Update puppet version, deprecate some older versions of OSes

wyardley:
This removes support for RedHat and derivatives v5 and Ubuntu 10, adds support for Debian 9, and changes the Puppet version requirement to the same one used in puppet-rabbitmq (">=4.8.0 < 6.0.0").

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -11,7 +11,6 @@
     {
       "operatingsystem": "RedHat",
       "operatingsystemrelease": [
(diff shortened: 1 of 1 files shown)


wyardley:
@ekohl I think supported version is not enforced, though it may affect which OS versions tests run against. So this shouldn't orphan someone using an unsupported version anyway.

Issue #312, opened 2017-09-08
test error when running locally


wyardley:
## Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: 5.1.0
- Ruby: 2.3.4p301
- Distribution: [running tests on Mac OS X]
- Module version: current git master

## How to reproduce (e.g Puppet code you use)
`bundle install`
`bundle exec rake test`

## What are you seeing
test failures [see below]. Tests don't appear to fail in Travis. Additionally, is the text "and install_options is a string" correct when it appears to be an array with a single string as the content? (`['--verbose']`)?

## What behaviour did you expect instead
tests to pass

## Output log
```
  1) Puppet::Type::Package::ProviderNpm when installing npm packages and install_options is a string passes the install_options to npm
     Failure/Error: npm('install', *options, package)
     
     Mocha::ExpectationError:
       unexpected invocation: Package[express](provider=npm).npm("install", "--global", "express")
       unsatisfied expectations:
       - expected exactly once, not yet invoked: Package[express](provider=npm).npm("install", "--global", "--verbose", "express")
       satisfied expectations:
       - allowed any number of times, not yet invoked: #<Puppet::Util::Feature:0x7faf8f99fd18>.root?(any_parameters)
       - allowed any number of times, not yet invoked: Puppet::Type::Package::ProviderNpm.command(:npm)
       - allowed any number of times, not yet invoked: Puppet::Type::Package::ProviderNpm.optional_commands(:npm)
     # ./lib/puppet/provider/package/npm.rb:76:in `install'
     # ./spec/unit/puppet/provider/package/npm_spec.rb:46:in `block (4 levels) in <top (required)>'

  2) Puppet::Type::Package::ProviderNpm when installing npm packages and install_options is a hash passes the install_options to npm
     Failure/Error: npm('install', *options, package)
     
     Mocha::ExpectationError:
       unexpected invocation: Package[express](provider=npm).npm("install", "--global", "express")
       unsatisfied expectations:
       - expected exactly once, not yet invoked: Package[express](provider=npm).npm("install", "--global", "--loglevel=error", "express")
       satisfied expectations:
       - allowed any number of times, not yet invoked: #<Puppet::Util::Feature:0x7faf8f99fd18>.root?(any_parameters)
       - allowed any number of times, not yet invoked: Puppet::Type::Package::ProviderNpm.command(:npm)
[...]

PR #315, opened 2017-09-14
Verify EPEL support in acceptance tests

ekohl:
Now CentOS 7.4 is out this pending statement can be removed.

https://lists.centos.org/pipermail/centos-announce/2017-September/022532.html

### DIFF
--- a/spec/acceptance/class_spec.rb
+++ b/spec/acceptance/class_spec.rb
@@ -32,7 +32,6 @@ describe 'nodejs class:', unless: UNSUPPORTED_PLATFORMS.include?(fact('os.family
     describe package('nodejs') do
       it { is_expected.to be_installed }
       it 'comes from the expected source' do
-        pending("This won't work until we have CentOS 7.4 because of dependency")
         pkg_output = shell(pkg_cmd)
         expect(pkg_output.stdout).to match 'epel'
       end


wyardley:
Acceptance test still seems to be seeing the source as nodesource.

Also, there's an unrelated mocha error that I've seen locally now showing up in the regular spec tests.

ekohl:
I will admit that I didn't run it locally, just sent it to Travis. Otherwise I'm not surprised this fails: the resource is ensured to be installed so most likely it needs to be removed before running the tests.

wyardley:
Also, in the Vagrant (vs. Docker) boxes, do they run a `yum update` when they come up, or would we need Puppetlabs to pack a new CentOS 7 box before these will work?

wyardley:
> Otherwise I'm not surprised this fails: the resource is ensured to be installed so most likely it needs to be removed before running the tests.
Yeah, we do something similar in puppet-nginx. Good point.
But also, I think the acceptance test box is still 7.3:
```
[root@centos-7-x64 ~]# cat /etc/redhat-release
CentOS Linux release 7.3.1611 (Core) 
```

ekohl:
> Also, in the Vagrant (vs. Docker) boxes, do they run a yum update when they come up, or would we need Puppetlabs to pack a new CentOS 7 box before these will work?

I expect they don't but don't have to either. There is only one CentOS 7 URL so it'll find the dependency and install it when needed. No need to update the full box though the docker image is already updated.

wyardley:
What do you think the cleanest way is to remove the package within the acceptance test? It's already running only on RedHat, so could I guess remove with yum / rpm, but maybe should use `puppet resource`?

ekohl:
```puppet resource``` is the easiest. We could add a ```before(:all)``` block to do that.

wyardley:
ps, I think the error in https://travis-ci.org/voxpupuli/puppet-nodejs/jobs/275380441 is the same as #312 
Not sure why that wasn't showing up earlier.

wyardley:
@ekohl have you had a chance to revisit this?

ekohl:
@wyardley no I haven't and I don't think I'll have time to in the near future.

wyardley:
No worries, I will see if I can work on it. We got some clarity on what the expectation is, so may just need to not try and check where the package is coming from.

PR #321, opened 2017-09-20
Allow for the macports provider to be overridden on macos

kizzie:
A vague pondering on how to fix this issue - not sure its 100% working or if my macvm is not happy about npm for other reasons. But any thoughts?

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -23,7 +23,12 @@ class nodejs(
   $repo_proxy_username                                 = $nodejs::params::repo_proxy_username,
   $repo_url_suffix                                     = $nodejs::params::repo_url_suffix,
   Array $use_flags                                     = $nodejs::params::use_flags,
+  String $darwin_package_provider                      = $nodejs::params::darwin_package_provider,
 ) inherits nodejs::params {
+
+  if $facts['osfamily'] == 'Darwin' {
+    Package { provider => $darwin_package_provider }
+  }
 
   if $manage_package_repo and !$repo_class {
     fail("${module_name}: The manage_package_repo parameter was set to true but no repo_class was provided.")
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -23,12 +23,8 @@ class nodejs(
   $repo_proxy_username                                 = $nodejs::params::repo_proxy_username,
   $repo_url_suffix                                     = $nodejs::params::repo_url_suffix,
   Array $use_flags                                     = $nodejs::params::use_flags,
-  String $darwin_package_provider                      = $nodejs::params::darwin_package_provider,
+  Optional[String] $package_provider                   = $nodejs::params::package_provider,
 ) inherits nodejs::params {
-
-  if $facts['osfamily'] == 'Darwin' {
-    Package { provider => $darwin_package_provider }
-  }
 
   if $manage_package_repo and !$repo_class {
     fail("${module_name}: The manage_package_repo parameter was set to true but no repo_class was provided.")
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -18,8 +18,9 @@ class nodejs::install {
     }
   }
 
+  Package { provider => $nodejs::package_provider }
   # nodejs
-  package { $nodejs::nodejs_package_name:
+  -> package { $nodejs::nodejs_package_name:
     ensure => $nodejs::nodejs_package_ensure,
     tag    => 'nodesource_repo',
   }
@@ -51,7 +52,7 @@ class nodejs::install {
   if $facts['osfamily'] == 'Darwin' {
     $root_npmrc_path = '/var/root'
   } else {
-    $root_npmrc_path = '/root'
+    $root_npmrc_path = $facts['root_home']
   }
 
   file { 'root_npmrc':
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -56,7 +56,7 @@ class nodejs::install {
 
   file { 'root_npmrc':
     ensure  => 'file',
-    path    => '/root/.npmrc',
+    path    => "${root_npmrc_path}/.npmrc",
     content => template('nodejs/npmrc.erb'),
     owner   => 'root',
     group   => '0',
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -48,6 +48,12 @@ class nodejs::install {
     }
   }
 
+  if $facts['osfamily'] == 'Darwin' {
+    $root_npmrc_path = '/var/root'
+  } else {
+    $root_npmrc_path = '/root'
+  }
+
   file { 'root_npmrc':
     ensure  => 'file',
     path    => '/root/.npmrc',
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -18,9 +18,8 @@ class nodejs::install {
     }
   }
 
-  if $nodejs::package_provider != undef {
-    Package { provider => $nodejs::package_provider }
-  }
+  Package { provider => $nodejs::package_provider }
+
   # nodejs
   package { $nodejs::nodejs_package_name:
     ensure => $nodejs::nodejs_package_ensure,
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -20,7 +20,7 @@ class nodejs::install {
 
   Package { provider => $nodejs::package_provider }
   # nodejs
-  -> package { $nodejs::nodejs_package_name:
+  package { $nodejs::nodejs_package_name:
     ensure => $nodejs::nodejs_package_ensure,
     tag    => 'nodesource_repo',
   }
@@ -49,11 +49,7 @@ class nodejs::install {
     }
   }
 
-  if $facts['osfamily'] == 'Darwin' {
-    $root_npmrc_path = '/var/root'
-  } else {
-    $root_npmrc_path = $facts['root_home']
-  }
+  $root_npmrc_path = $facts['root_home']
 
   file { 'root_npmrc':
     ensure  => 'file',
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -18,7 +18,9 @@ class nodejs::install {
     }
   }
 
-  Package { provider => $nodejs::package_provider }
+  if $nodejs::package_provider != undef {
+    Package { provider => $nodejs::package_provider }
+  }
   # nodejs
   package { $nodejs::nodejs_package_name:
     ensure => $nodejs::nodejs_package_ensure,
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -50,11 +50,9 @@ class nodejs::install {
     }
   }
 
-  $root_npmrc_path = $facts['root_home']
-
   file { 'root_npmrc':
     ensure  => 'file',
-    path    => "${root_npmrc_path}/.npmrc",
+    path    => "${facts['root_home']}/.npmrc",
     content => template('nodejs/npmrc.erb'),
     owner   => 'root',
     group   => '0',
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -55,7 +55,7 @@ class nodejs::params {
         $repo_class                = '::nodejs::repo::nodesource'
       }
 
-      $package_provider          = 'apt'
+      $package_provider          = undef
     }
     'RedHat': {
       if $facts['os']['release']['major'] =~ /^[67]$/ {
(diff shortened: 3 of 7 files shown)


kizzie:
Hmm, so the next issue with this is that the /root/.npmrc file can't be created for npm as that user doesn't exist for macos

kizzie:
There we go, working when you pass in 

` class { 'nodejs' : 
  mac_provider => 'homebrew'
} `

if nothing else

manifests/init.pp
```diff
@@ -24,8 +24,20 @@
   $repo_proxy_username                                 = $nodejs::params::repo_proxy_username,
   $repo_url_suffix                                     = $nodejs::params::repo_url_suffix,
   Array $use_flags                                     = $nodejs::params::use_flags,
+  $mac_provider                                        = undef,
```
juniorsysadmin: This isn't the right way to do it. For consistency, the best approach is to create a new variable in params.pp `$darwin_package_provider = 'macports'` then do `$darwin_package_provider = $nodejs::params::darwin_package_provider` in init.pp

manifests/install.pp
```diff
@@ -60,9 +60,15 @@
     }
   }
 
+  if $::osfamily == 'Darwin' {
+    $path = '/var/root'
```
juniorsysadmin: Given `$path` is a banned keyword now I think, consider something like `$root_npmrc_path = '/var/root/.npmrc'` instead

manifests/init.pp
kizzie: Oooh good point, I was having issues with the provider not overriding when the package statement was in the params as well, but actually that could be moved to init and the default still set in params.

manifests/install.pp
kizzie: Oh dear, that's going to cause some fun for me (just starting to move things from puppet 3 to 4 and must have missed this one!)

kizzie:
I'll need to do some reading up on how to get travis to add tests for osx (I'm not sure beaker will be happy at all but there must be a way!)

juniorsysadmin:
@kizzie I'm okay with merging minus the travis.yml changes, leaving matching rspec-puppet tests for another Pull Request. 

> I'll need to do some reading up on how to get travis to add tests for osx

I wouldn't bother with this for now.

kizzie:
Okey dokey, shall remove the travis bits, as I never did get time to look at it - the rspec ones should at least be easier than trying to force beaker into working forit  :)

manifests/init.pp
```diff
@@ -24,8 +24,13 @@
   $repo_proxy_username                                 = $nodejs::params::repo_proxy_username,
   $repo_url_suffix                                     = $nodejs::params::repo_url_suffix,
   Array $use_flags                                     = $nodejs::params::use_flags,
+  $darwin_package_provider                             = $nodejs::params::darwin_package_provider,
 ) inherits nodejs::params {
 
+  if $::osfamily == 'Darwin' {
```
bastelfreak: please use the new facts hash here:
$facts['os']['family']

manifests/init.pp
```diff
@@ -24,8 +24,13 @@
   $repo_proxy_username                                 = $nodejs::params::repo_proxy_username,
   $repo_url_suffix                                     = $nodejs::params::repo_url_suffix,
   Array $use_flags                                     = $nodejs::params::use_flags,
+  $darwin_package_provider                             = $nodejs::params::darwin_package_provider,
```
bastelfreak: please add a puppet4 datatype for this new parameter

kizzie:
Do you have a preferred way to squash all the commits? Its starting to look a bit of a mess :)

bastelfreak:
Personally I prefer a history with small commits, so this is in general okay. Can you rebase and get rid of 2c1ca6772d5ff6610bc93cdb08b6b1923b6736b1 and 96e53dc65c257b2601183652948bb1848ca3028f?
 
0511a52142343c51ed59ca0a92c863846a2d6725, 0276ab17728d649cdfd42138fbcd3ae936f47086 and 5911eb386f5d62ae5e3948175abaa380dde3e1f7
can be squashed into f290be1950fac5c3aed206843e0adabdd58b0d00

bastelfreak:
Can you check the email address used in the first 6 commits? The used email address isn't associated with your github account.

kizzie:
Oooh work laptop vs home computer. That's awkward. And annoyingly identifiable. Ho hum!

kizzie:
So my initial guess was correct, the rspec tests don't like $facts['os']['family'] for anything but ubuntu based systems - it's happy with $facts['osfamily'] though - Not sure whether you want me to go through and work out how to change all the tests to work, or whether to use the older style fact hash?

bastelfreak:
$facts['os']['family'] is probably not correctly mocked :(. Using the older $facts['osfamily']  is fine as well. It won't disappear soon.

kizzie:
All fixed up - with some added tests and a bit in the read me :)

manifests/init.pp
```diff
@@ -24,8 +24,13 @@
   $repo_proxy_username                                 = $nodejs::params::repo_proxy_username,
   $repo_url_suffix                                     = $nodejs::params::repo_url_suffix,
   Array $use_flags                                     = $nodejs::params::use_flags,
+  String $darwin_package_provider                      = $nodejs::params::darwin_package_provider,
```
ekohl: Would it make sense to just name it ```package_provider``` and only override it in ```install.pp```? The other distros  can default to undef.

manifests/init.pp
```diff
@@ -24,8 +24,13 @@
...
 ) inherits nodejs::params {
 
+  if $facts['osfamily'] == 'Darwin' {
+    Package { provider => $darwin_package_provider }
```
ekohl: I wouldn't mind seeing this in ```install.pp``` instead.

manifests/install.pp
```diff
@@ -60,9 +60,15 @@
     }
   }
 
+  if $facts['osfamily'] == 'Darwin' {
+    $root_npmrc_path = '/var/root'
```
ekohl: Is the ```root_home``` fact useful here?

manifests/init.pp
juniorsysadmin: Hmmm..maybe. Doing it the current way makes it more explicit that Darwin is it's own special snowflake rather than partially supporting your choice of `package_provider` regardless of os.

ekohl: I don't like special snowflakes :)

juniorsysadmin: > I don't like special snowflakes :)
Realistically noone is going to have any reason to change `package_provider` for any of the other operating systems supported by this module.

kizzie: Darwin is always its own special snowflake for anything puppet related. It always acts slightly differently (grumble... grump... etc.) We would need to pass the package_provider variable over to install and then use it there but it shouldn't change functionality in anyway - just makes it a bit harder to see where its going if you're using init. I don't mind one way or the other

manifests/install.pp
kizzie: Is that a custom fact from somewhere? I'm not seeing it in any of the facter outputs for ubuntu/macos - it is probably just me though!

bastelfreak: @ekohl good catch!

-----

@kizzie https://github.com/puppetlabs/puppetlabs-stdlib#root_home it comes from stdlib

kizzie: Ooh I had not heard of that one - shall give it a try later!

manifests/init.pp
ekohl: > Would it make sense to just name it package_provider and only override it in install.pp? The other distros can default to undef.

Apologies, I meant ```params.pp``` instead of ```install.p```.

> Realistically noone is going to have any reason to change package_provider for any of the other operating systems supported by this module.

From [pep 20](https://www.python.org/dev/peps/pep-0020/):

> Special cases aren't special enough to break the rules.

I don't see how renaming it to ```Optional[String] $package_provider``` and providing sane defaults in ```params.pp``` would be a difference for Darwin users but it would generalize it neatly.

juniorsysadmin:
@kizzie Did you still need this?

kizzie:
Oh yes, we switched to doing something else internally (because even homebrew support wasn't giving us the version we were after) but will try and fix this up this week for your comments

manifests/init.pp
kizzie: Have generalised this so passing in the package_provider for everything which can be overidden as and when!

manifests/install.pp
kizzie: So I've put this in but it involves changing the test as its usually not being set to anything for the tests and so being resolved as /.npmrc

kizzie:
rebased to master now so hopefully should all be working!

kizzie:
There we go. Different morning, different realisation that master may have updated. Hopefully all in order now.

manifests/params.pp
```diff
@@ -57,6 +57,8 @@
         $npm_path                  = '/usr/bin/npm'
         $repo_class                = '::nodejs::repo::nodesource'
       }
+
+      $package_provider          = 'apt'
```
ekohl: I'd prefer to set this and others (yum, zypper etc) to undef unless it's non-standard (macports? chocolatey?).

kizzie: Changed it over

manifests/install.pp
```diff
@@ -18,7 +18,9 @@
     }
   }
 
-  Package { provider => $nodejs::package_provider }
+  if $nodejs::package_provider != undef {
```
ekohl: Technically this is not needed. When ```undef``` it'll fall back on the default value from puppet.

spec/classes/nodejs_spec.rb
```diff
@@ -18,7 +18,7 @@
       it 'the file resource root_npmrc should be in the catalog' do
         is_expected.to contain_file('root_npmrc').with(
           'ensure' => 'file',
-          'path'    => '/root/.npmrc',
+          'path'    => '/.npmrc',
```
ekohl: This is unexpected. We should fix a value for that fact. Could you add it to ```spec/default_facts.yml```? I'll submit a PR to modulesync_config to ensure it's not overridden.

spec/classes/nodejs_spec.rb
```diff
@@ -920,6 +920,16 @@
       }
     end
 
+    it 'the file resource root_npmrc should use /var/root for the path' do
+      is_expected.to contain_file('root_npmrc').with(
+        'ensure' => 'file',
+        'path'    => '/.npmrc',
```
ekohl: This doesn't match the description. Maybe this should be left out since it's already tested above.

kizzie: We could either leave it out or override the fact for this test to be closer to the default that Darwin should come up with?

kizzie:
I have no idea why the build is failing to find bundle, any ideas?

.travis.yml
```diff
@@ -7,6 +7,7 @@ before_install:
   - bundle -v
   - rm Gemfile.lock || true
   - gem update --system
+  - gem pristine bundler
```
juniorsysadmin: Drop this line - after a rebase to master you should have a green build

spec/default_facts.yml
```diff
@@ -12,3 +12,4 @@ concat_basedir: "/tmp"
 ipaddress: "172.16.254.254"
 is_pe: false
 macaddress: "AA:AA:AA:AA:AA:AA"
+root_home: /root
```
juniorsysadmin: Add this to `default_module_facts.yaml` as well

kizzie: Done

ekohl: If it's in ```default_module_facts.yaml``` then can't it be removed here? This file is modulesynced while the other isn't.

manifests/install.pp
```diff
@@ -48,9 +50,11 @@
     }
   }
 
+  $root_npmrc_path = $facts['root_home']
+
   file { 'root_npmrc':
     ensure  => 'file',
-    path    => '/root/.npmrc',
+    path    => "${root_npmrc_path}/.npmrc",
```
ekohl: You can use ```"${facts['root_home']}/.npmrc"``` to remove the additional variable.

spec/default_facts.yml
ekohl: Then I'm not that clear of mind yet. Just leave it in and if it's an issue in the modulesync we'll fix it then.

juniorsysadmin:
@kizzie Thanks for this PR, I have just squash commits and merged.

ekohl:
Thanks for addressing all of my many comments @kizzie!

Issue #324, opened 2017-09-26
Don't restrict which Fedora versions can use the NodeSource repository


juniorsysadmin:
https://github.com/voxpupuli/puppet-nodejs/blob/master/manifests/repo/nodesource.pp#L31 contains a list of versions. They should be removed and any Fedora version should be allowed.

PR #325, opened 2017-10-01
Allow all Fedora versions to use NodeSource

juniorsysadmin:
Don't restrict what versions can use the NodeSource repository


juniorsysadmin:
Fixes #324

bastelfreak:
Thanks for the PR!

PR #327, opened 2017-10-17
Deprecate EOL Fedora versions

ghoneycutt:


ghoneycutt:
Getting reacquainted with the code and noticed some ancient Fedora versions. By removing these EOL versions we can speed up the testing and not be tied to them.

bastelfreak:
Thanks @ghoneycutt!

PR #334, opened 2017-11-12
Set default value for $repo_pin to undef

### LINKED ISSUE #333: apt pin error

- Puppet: 5.2.0
- Ruby: 2.3.1p112
- Distribution: Ubuntu 16.04
- Module version: 4.0.0

class { 'nodejs':
        repo_url_suffix => '8.x',
}

## What are you seeing
Error: Evaluation Error: Error while evaluating a Resource Statement, Apt::Source[nodesource]: parameter 'pin' expects a value of type Undef, Hash, Numeric, or String, got Boolean  at /srv/maverick/software/maverick/manifests/puppet-modules/nodejs/manifests/repo/nodesource/apt.pp:14 on node maverick-ubuntuvm.siliconalchemy.net
```

### PR DESCRIPTION
**Pull request by juniorsysadmin**
https://github.com/puppetlabs/puppetlabs-apt/pull/711 introduced data types in the apt module, and false is no longer a valid value for the apt::source $pin parameter

Fixes #333

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -5,7 +5,7 @@ class nodejs::params {
   $nodejs_package_ensure       = 'present'
   $repo_enable_src             = false
   $repo_ensure                 = 'present'
-  $repo_pin                    = false
+  $repo_pin                    = undef
   $repo_priority               = 'absent'
   $repo_proxy                  = 'absent'
   $repo_proxy_password         = 'absent'
--- a/spec/classes/nodejs_spec.rb
+++ b/spec/classes/nodejs_spec.rb
@@ -124,11 +124,11 @@ describe 'nodejs', type: :class do
 
         context 'and repo_pin not set' do
           let :params do
-            default_params.merge!(repo_pin: 'false')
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: manifests/params.pp [hunk 1]
```diff
@@ -5,7 +5,7 @@
   $nodejs_package_ensure       = 'present'
   $repo_enable_src             = false
   $repo_ensure                 = 'present'
-  $repo_pin                    = false
+  $repo_pin                    = undef
```
juniorsysadmin: Not making this a data type for now as want to keep the diff small

juniorsysadmin:
This change should continue to work with earlier 2/3 versions of soft dependency apt

spec/classes/nodejs_spec.rb
```diff
@@ -122,13 +122,13 @@
           end
         end
 
-        context 'and repo_pin not set' do
```
alexjfisher: The original statement is probably more accurate.  If you set something to `undef`, you're not actually setting it.  The parameter default gets used (which happens to be `undef`).

PR #342, opened 2017-11-18
Enhance distro support

fnoop:
Add explicit Debian 9 (stretch) support, and add support for Ubilinux.

Ubilinux 4 (that supports all variants of the Up boards) is a derivative of Debian 9.  However it reports dolcetto as the lsbdistcodename instead of stretch so the nodejs repo fails.  This PR adds support for a $nodejs::repo_release parameter that allows the distro code name to be specified rather than blindly taking it from lsbdistcodename.

eg.:
```
    class { 'nodejs':
        repo_url_suffix => '8.x',
        repo_release    => 'stretch',
        nodejs_package_ensure => latest,
    }
```

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -23,6 +23,7 @@ class nodejs(
   $repo_proxy_password                                 = $nodejs::params::repo_proxy_password,
   $repo_proxy_username                                 = $nodejs::params::repo_proxy_username,
   $repo_url_suffix                                     = $nodejs::params::repo_url_suffix,
+  $repo_release                                        = $nodejs::params::repo_release,
   Array $use_flags                                     = $nodejs::params::use_flags,
 ) inherits nodejs::params {
 
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -23,7 +23,7 @@ class nodejs(
   $repo_proxy_password                                 = $nodejs::params::repo_proxy_password,
   $repo_proxy_username                                 = $nodejs::params::repo_proxy_username,
   $repo_url_suffix                                     = $nodejs::params::repo_url_suffix,
-  $repo_release                                        = $nodejs::params::repo_release,
+  Optional[String] $repo_release                       = $nodejs::params::repo_release,
   Array $use_flags                                     = $nodejs::params::use_flags,
 ) inherits nodejs::params {
 
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -25,7 +25,7 @@ class nodejs::params {
       if $::operatingsystemrelease =~ /^6\.(\d+)/ {
         fail("The ${module_name} module is not supported on Debian Squeeze.")
       }
-      elsif $::operatingsystemrelease =~ /^[78]\.(\d+)/ {
+      elsif $::operatingsystemrelease =~ /^[789]\.(\d+)/ {
         $legacy_debian_symlinks    = false
         $manage_package_repo       = true
         $nodejs_debug_package_name = 'nodejs-dbg'
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -11,6 +11,7 @@ class nodejs::params {
   $repo_proxy_password         = 'absent'
   $repo_proxy_username         = 'absent'
   $repo_url_suffix             = '0.10'
+  $repo_release                = $::lsbdistcodename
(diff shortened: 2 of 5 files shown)


manifests/params.pp
```diff
@@ -25,7 +26,7 @@
       if $::operatingsystemrelease =~ /^6\.(\d+)/ {
         fail("The ${module_name} module is not supported on Debian Squeeze.")
       }
-      elsif $::operatingsystemrelease =~ /^[78]\.(\d+)/ {
+      elsif $::operatingsystemrelease =~ /^[789]\.(\d+)/ {
```
ekohl: Wouldn't ```versioncmp($::operatingsystemmajrelease, '7') >= 0``` be better and more future proof?

fnoop: I think that would incorrectly catch the Ubuntu versions as well wouldn't it?

fnoop: This will likely break for the next version of Debian (10) as it will start to conflict with Ubuntu 10.x, logic should be broken off separately for Debian and Ubuntu.

ekohl: You're right. Early morning code reviews aren't the best idea :P

manifests/repo/nodesource/apt.pp
```diff
@@ -5,11 +5,18 @@
   $ensure     = $nodejs::repo::nodesource::ensure
   $pin        = $nodejs::repo::nodesource::pin
   $url_suffix = $nodejs::repo::nodesource::url_suffix
+  $release    = $nodejs::repo::nodesource::release
 
   ensure_packages(['apt-transport-https', 'ca-certificates'])
 
   include ::apt
 
+  if empty($release) {
```
ekohl: Would ```pick($release, $::lsbdistcodename)``` be better here or is this another case of me missing anything? :)

fnoop: Didn't know about pick, you're right!

fnoop:
Note: the failing travis checks appears to be because of missing nodejs-dev package, but the managed nginx repo doesn't seem to provide this package - headers are included in the main nodejs package by the look of it.

manifests/params.pp
```diff
@@ -11,6 +11,7 @@
   $repo_proxy_password         = 'absent'
   $repo_proxy_username         = 'absent'
   $repo_url_suffix             = '0.10'
+  $repo_release                = undef
```
juniorsysadmin: It's clearer if you default to $::lsbdistcodename here, no?

manifests/init.pp
```diff
@@ -23,6 +23,7 @@
   $repo_proxy_password                                 = $nodejs::params::repo_proxy_password,
   $repo_proxy_username                                 = $nodejs::params::repo_proxy_username,
   $repo_url_suffix                                     = $nodejs::params::repo_url_suffix,
+  $repo_release                                        = $nodejs::params::repo_release,
```
juniorsysadmin: Please add an entry to the README for this

manifests/params.pp
fnoop: As per https://github.com/voxpupuli/puppet-nginx/pull/1154#discussion_r151865121, changing to pass $repo_release through unaltered and let apt module deal with undef release.

manifests/init.pp
fnoop: Added README entry
https://github.com/voxpupuli/puppet-nodejs/pull/342/commits/40f96bb2035b578ec1602581c822c170e7028e6a

bastelfreak: Can you add a datatype for the new param?

manifests/params.pp
fnoop: Added datatype to new parameter, and pass it straight through apt module in commits https://github.com/voxpupuli/puppet-nodejs/pull/342/commits/fdb655df9c5ddb47922c112c9f65cc757da8f473  https://github.com/voxpupuli/puppet-nodejs/pull/342/commits/63c3d3c7812f002d850f9658e7eae7f33a86fa58

juniorsysadmin:
@fnoop Are you now able rebase on top of master to fix the merge conflicts? (#343 was merged, which had some of your changes)

juniorsysadmin:
Have stolen the relevant bits in here and put them in #349

PR #343, opened 2017-11-29
Remove EOL operating systems, Legacy facts.

juniorsysadmin:
- Update the `repo_url_suffix` table in the README, dropping EOL
operating systems and adding new ones
- Drop explicit handling of Debian 6 and 7 (EOL)
- Drop explicit handling of Ubuntu 10.04 and 12.04 (EOL)
- Drop explicit handling of EOL Fedora versions
- Drop explicit handling of RHEL5 (EOL)
- Add explicit handling of Debian 9
- Replace most Legacy facts with Modern facts using the new facts hash
syntax. This explicitly drops support for Facter versions less than 3,
which are not supported under Puppet4/5 anyway.
- Drop explicit support for Amazon Linux versions less than 2015.03.
Amazon Linux is rolling distribution, and earlier versions did not have
Facter 3
- Use the Modern system32 fact

Also:
- Remove the old Puppetlabs footer from the README Reference CONTRIBUTING.md instead
- Perform minor doc fixes
- Use less granular error handling

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,7 +75,6 @@ When a separate npm package exists (natively or via EPEL) the Node.js developmen
 package also needs to be installed as it is a dependency for npm.
 
 Install Node.js and npm using the native packages provided by the distribution:
-(Only applicable for Ubuntu 12.04/14.04 and Fedora operating systems):
 
 ```puppet
 class { '::nodejs':
@@ -470,13 +469,11 @@ the NodeSource URL structure - NodeSource might remove old versions (such as
 0.10 and 0.12) or add new ones (such as 8.x) at any time.
 
 The following are ``repo_url_suffix`` values that reflect NodeSource versions
-that were available on 2017-11-14:
+that were available on 2017-11-29:
 
-* Debian 7 (Wheezy) ```0.10``` ```0.12``` ```4.x``` ```5.x``` ```6.x```
 * Debian 8 (Jessie) ```0.10``` ```0.12``` ```4.x``` ```5.x``` ```6.x``` ```7.x``` ```8.x``` ```9.x```
 * Debian 9 (Stretch) ```4.x``` ```6.x``` ```7.x``` ```8.x``` ```9.x```
 * Debian (Sid) ```0.10``` ```0.12``` ```4.x``` ```5.x``` ```6.x``` ```7.x``` ```8.x``` ```9.x```
-* Ubuntu 12.04 (Precise) ```0.10``` ```0.12``` ```4.x``` ```5.x``` ```6.x```
 * Ubuntu 14.04 (Trusty) ```0.10``` ```0.12``` ```4.x``` ```5.x``` ```6.x``` ```7.x``` ```8.x``` ```9.x```
 * Ubuntu 16.04 (Xenial) ```0.10``` ```0.12``` ```4.x``` ```5.x``` ```6.x``` ```7.x``` ```8.x``` ```9.x```
 * Ubuntu 16.10 (Yakkety) ```0.12``` ```4.x``` ```6.x``` ```7.x``` ```8.x```
@@ -487,6 +484,7 @@ that were available on 2017-11-14:
 * Amazon Linux - See RHEL/CentOS 7
 * Fedora 25 ```4.x``` ```6.x``` ```7.x``` ```8.x``` ```9.x```
 * Fedora 26 ```6.x``` ```8.x``` ```9.x```
+* Fedora 27 ```8.x``` ```9.x```
 
 #### `use_flags`
 
@@ -497,23 +495,22 @@ The USE flags to use for the Node.js package on Gentoo systems. Defaults to
 
 This module has received limited testing on:
 
-* CentOS/RHEL 5/6/7
-* Debian 7
-* Ubuntu 10.04/12.04/14.04
+* CentOS/RHEL 6/7
+* Debian 8
+* Ubuntu 14.04
 
 The following platforms should also work, but have not been tested:
 
 * Amazon Linux
 * Archlinux
 * Darwin
-* Debian 8
+* Debian 9
+* Fedora
 * FreeBSD
 * Gentoo
 * OpenBSD
 * OpenSuse/SLES
 * Windows
-
-This module is not supported on Debian Squeeze.
 
 ### Module dependencies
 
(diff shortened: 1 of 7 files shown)


juniorsysadmin:
I haven't replaced `$::lsbdistcodename`
`$facts['os'['distro']['codename']` (to replace `$::lsbdistcodename` in nodesource/apt.pp) is not handled by rspec-puppet-facts, so will need mocked values.

manifests/params.pp
```diff
@@ -37,43 +32,7 @@
...
-        $nodejs_dev_package_name   = 'nodejs-dev'
-        $nodejs_dev_package_ensure = 'absent'
-        $nodejs_package_name       = 'nodejs'
-        $npm_package_ensure        = 'absent'
-        $npm_package_name          = 'npm'
-        $npm_path                  = '/usr/bin/npm'
-        $repo_class                = '::nodejs::repo::nodesource'
-      }
-      elsif $::operatingsystemrelease =~ /^16.04$/ {
+      elsif $facts['os']['release']['full'] =~ /^14.04$|^16.04$/ {
```
ekohl: How about ```/^1[46]\.04$/``` for the regex?

manifests/repo/nodesource.pp
```diff
@@ -9,32 +9,26 @@
...
+  case $facts['os']['family'] {
     'RedHat': {
-      if $::operatingsystemrelease =~ /^5\.(\d+)/ {
-        include ::epel
-        $dist_version  = '5'
-        $name_string   = 'Enterprise Linux 5'
-      }
-
-      elsif $::operatingsystemrelease =~ /^6\.(\d+)/ {
+      if $facts['os']['release']['full'] =~ /^6\.(\d+)/ {
```
ekohl: The ```major``` should be there in facts too and doesn't need a regex. You can even simplify the later code to use those facts in ```$dist_version``` and ```$name_string```.

manifests/repo/nodesource.pp
```diff
@@ -59,52 +53,19 @@
       class { '::nodejs::repo::nodesource::yum': }
       contain '::nodejs::repo::nodesource::yum'
 
-      if $::operatingsystemrelease =~ /^5\.(\d+)/ {
-        # On EL 5, EPEL needs to be applied first
-        Class['::epel'] -> Class['::nodejs::repo::nodesource::yum']
-      }
-
     }
     'Linux': {
```
ekohl: I think you can just remove the whole Linux block and allow it to fall back to the generic default.

spec/classes/nodejs_spec.rb
```diff
@@ -208,10 +208,9 @@
           }
         end
 
-        if facts[:osfamily] == 'Debian' && (
-             facts[:operatingsystemrelease] == '10.04' ||
-             %w[7 8].include?(facts[:operatingsystemmajrelease])
-        )
+        if facts[:os]['family'] == 'Debian' &&
```
ekohl: I'd extract this into a separate variable since it's reused a lot.

spec/classes/nodejs_spec.rb
```diff
@@ -326,25 +322,11 @@
...
-      }
-    end
-
-    it do
-      expect { catalogue }.to raise_error(Puppet::Error, %r{The nodejs module is not supported on Fedora 18.})
-    end
-  end
-
-  ['5.0', '6.0', '7.0', '25'].each do |operatingsystemrelease|
+  ['6.0', '7.0', '27'].each do |operatingsystemrelease|
```
ekohl: This and a lot below should be in ```on_supported_os``` too. Could we reuse that instead of duplicating facts?

manifests/params.pp
juniorsysadmin: Done

manifests/repo/nodesource.pp
juniorsysadmin: Done

manifests/repo/nodesource.pp
juniorsysadmin: Done

spec/classes/nodejs_spec.rb
juniorsysadmin: Hmm maybe. We also would need to filter CentOS/OracleLinux/Scientific Linux as we only one EL6/7 types tested once.

README.md
```diff
@@ -497,34 +495,30 @@ The USE flags to use for the Node.js package on Gentoo systems. Defaults to
...
 
 This modules uses `puppetlabs-apt` for the management of the NodeSource
 repository. If using an operating system of the Debian-based family, you will
 need to ensure that `puppetlabs-apt` version 2.x or above is installed.
 
-If using CentoOS/RHEL 5, you will need to ensure that the `stahnma-epel`
-module is installed.
-
-If using CentoOS/RHEL 5/6/7 and you wish to install Node.js from EPEL rather
+If using CentoOS/RHEL 6/7 and you wish to install Node.js from EPEL rather
```
ekohl: CentOS

manifests/params.pp
```diff
@@ -13,19 +13,14 @@
...
   }
 
-  case $::osfamily {
+  case $facts['os']['family'] {
     'Debian': {
-      if $::operatingsystemrelease =~ /^6\.(\d+)/ {
-        fail("The ${module_name} module is not supported on Debian Squeeze.")
-      }
-      elsif $::operatingsystemrelease =~ /^[78]\.(\d+)/ {
+      if $facts['os']['release']['full'] =~ /^[89]\.(\d+)/ {
```
ekohl: Should this use ```['major'] =~ /^[89]$/```

manifests/params.pp
```diff
@@ -102,7 +61,7 @@
     'RedHat': {
       $legacy_debian_symlinks      = false
 
-      if $::operatingsystemrelease =~ /^[5-7]\.(\d+)/ {
+      if $facts['os']['release']['full'] =~ /^[6-7]\.(\d+)/ {
```
ekohl: Also use major?

manifests/params.pp
```diff
@@ -227,43 +185,10 @@
       $repo_class                = undef
     }
     'Linux': {
```
ekohl: I think this adds little over the default failure

juniorsysadmin: Have dropped it.

manifests/params.pp
juniorsysadmin: Done

manifests/params.pp
juniorsysadmin: Done

README.md
juniorsysadmin: Fixed

PR #348, opened 2017-12-25
Use puppetlabs-apt to handle apt-transport-https

### LINKED ISSUE #345: Rely on puppetlabs-apt 4.4.0 for apt-transport-https installation

juniorsysadmin:
puppetlabs-apt 4.4.0 now handles installation of apt-transport-https for
a selection of Debian families: wheezy, jessie, stretch, trusty, xenial.

This change should also ensure that this module works on Debian 10, which
no longer has the apt-transport-https package.

Fixes #345

### DIFF
--- a/README.md
+++ b/README.md
@@ -516,7 +516,7 @@ The following platforms should also work, but have not been tested:
 
 This modules uses `puppetlabs-apt` for the management of the NodeSource
 repository. If using an operating system of the Debian-based family, you will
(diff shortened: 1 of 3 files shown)


juniorsysadmin:
Breaking change as if you previously were using a Debian-flavour, but not using wheezy, jessie, stretch, trusty, xenial, you now need to handle installation of apt-transport-https yourself.

PR #350, opened 2017-12-25
[DO NOT MERGE] Update travis config

ekohl:
This is just to test if this would work. It belongs in a modulesync.


juniorsysadmin:
#351

PR #352, opened 2017-12-26
Add info for upgrades, repository priorities

### LINKED ISSUE #322: Docs: Add info important for upgrades, repository priorities
Add more information to the README important for working upgrades.
- `nodejs_package_ensure` defaults to present - changing `repo_url_suffix` probably won't always result in a new version being installed. You might need to change the value to say `nodejs_package_ensure => "6.*"`
- Use `nodejs::repo_pin` or `nodejs::repo_priority` to ensure that the NodeSource repository takes precedence when the native NodeJS version is the same as the one available in NodeSource.

juniorsysadmin:
Update the README to provide more information about upgrades and forcing the use of NodeSource packages over native ones.

Fixes #322

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,29 @@ class { '::nodejs':
 }
 ```
 
+### Upgrades
+
+The parameter `nodejs_package_ensure` defaults to `present`. Changing the
+`repo_url_suffix` will not result in a new version being installed. Changing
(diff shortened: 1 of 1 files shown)


bastelfreak:
Thanks for the PR @juniorsysadmin!

PR #357, opened 2018-01-07
Set default version to 8.x

### LINKED ISSUE #356: Default to a newer Nodejs version (8.x)
As noticed in #355 this module defaults to installing 0.10.x which is outdated and no longer gets security updates. We should install something more recent by default - perhaps 8.x given that it is LTS.
At the moment, with 0.10.x set it will fail on Debian 9 (Stretch), Ubuntu 17.10 (Artful) and Fedora with the default values.

juniorsysadmin:
Node.js version 8.x is the current LTS version.

- Also remove references to RHEL5 since support has been dropped
- Also change some unit tests to reflect current versions and actual
  use cases

Fixes #356

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,23 +51,18 @@ To install Node.js and npm (using the NodeSource repository if possible):
 class { 'nodejs': }
 ```
 
-If you wish to install a Node.js 0.12.x release from the NodeSource repository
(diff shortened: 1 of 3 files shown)


bastelfreak:
Thanks!

Issue #361, opened 2018-04-28
Ubuntu 18.04 gives warning


fnoop:
## Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: 5.5.1
- Ruby: 2.5.1
- Distribution: Ubuntu 18.04 (bionic beaver)
- Module version: latest

## What are you seeing
`Warning: Scope(Class[Nodejs::Params]): The nodejs module might not work on Ubuntu 18.04. Sensible defaults will be attempted.`

## What behaviour did you expect instead
No warning

## Any additional information you'd like to impart
It looks like install works because the nodesource repo now exists for bionic.

fnoop:
Note: Ubuntu 18.04 is now the latest LTS distro, important that it is fully supported now.

fnoop:
@bastelfreak - Sorry acceptance tests completely baffle me, I have no idea even how to run them.  I see in spec/acceptance/nodesets/ I could add an xml file for 18.04, but the files say:
```
# This file is managed via modulesync
# https://github.com/voxpupuli/modulesync
# https://github.com/voxpupuli/modulesync_config
```
So I'm guessing adding it manually is the wrong thing to do.

If you could point me in the right direction, be happy to have a go though?

PR #362, opened 2018-04-28
Add support for ubuntu bionic 18.04, #361

fnoop:
A tiny fix to support Ubuntu 18.04 LTS (Bionic Beaver)

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -32,7 +32,7 @@ class nodejs::params {
         $npm_path                  = '/usr/bin/npm'
         $repo_class                = '::nodejs::repo::nodesource'
       }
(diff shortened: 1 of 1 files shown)


fnoop:
#361

fnoop:
Yes tested locally on 18.04, works fine - it's literally a one character fix.  The OS detection/support isn't very robust and will need to be changed for the next Ubuntu release, but works for 18.04.
No parameters, it will behave as before as with other OS.

bastelfreak:
https://github.com/camptocamp/facterdb/pull/82#event-1600066178 spec tests should work soonish

mpdude:
Could you please release a 5.0.1 to include this?

Issue #305, opened 2017-07-24
install/run local package.json as non-root user does not work (access denied)


jblom:
## Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: 4.8.0
- Ruby: 2.1.9p490
- Distribution: within Docker Centos 7 image
- Module version: latest

## How to reproduce (e.g Puppet code you use)

nodejs::npm { "install ${service_name} npm packages as user ${service_user}":
		ensure => 'present',
		user => $service_user,
		source => "/home/${service_user}/${service_name}/src/static/package.json",
		target => "/home/${service_user}/${service_name}/src/static",
		require => Class['nodejs'],
	}

## What are you seeing
The process is trying to access /root/.npm while it is running as a different user:

e.g. part of the log: "npm ERR!  { Error: EACCES: permission denied, mkdir '/root/.npm'"

## What behaviour did you expect instead

I expected that the process was not trying to use anything inside the root dir. When I run npm install in the source directory as that $service_user it works.

## Output log

Notice: /Stage[main]/Rd_infra_service::Implementation::Media_suite/Rd_infra_service::Resource::Service[wp5_mediasuite]/Rd_infra_service::Resource::Webclient[wp5_mediasuite]/Nodejs::Npm[install wp5_mediasuite npm packages as user clariah]/Exec[npm_install_install wp5_mediasuite npm packages as user clariah]/returns: npm ERR! addLocal Could not install /home/clariah/wp5_mediasuite/src/static/package.json
Notice: /Stage[main]/Rd_infra_service::Implementation::Media_suite/Rd_infra_service::Resource::Service[wp5_mediasuite]/Rd_infra_service::Resource::Webclient[wp5_mediasuite]/Nodejs::Npm[install wp5_mediasuite npm packages as user clariah]/Exec[npm_install_install wp5_mediasuite npm packages as user clariah]/returns: npm ERR! Linux 4.9.36-moby
Notice: /Stage[main]/Rd_infra_service::Implementation::Media_suite/Rd_infra_service::Resource::Service[wp5_mediasuite]/Rd_infra_service::Resource::Webclient[wp5_mediasuite]/Nodejs::Npm[install wp5_mediasuite npm packages as user clariah]/Exec[npm_install_install wp5_mediasuite npm packages as user clariah]/returns: npm ERR! argv "/usr/bin/node" "/usr/bin/npm" "install" "/home/clariah/wp5_mediasuite/src/static/package.json"
[...]

PR #363, opened 2018-05-02
Fix default install as non-root user, #305

fnoop:
npm.pp currently sets /root as the environment HOMEDIR.  This breaks using nodejs::npm as a non-root user, as per #305.  nodejs::npm::home_dir parameter can be set to override this, but is not intuitive, documented or automatic.

Instead, unset $home_dir parameter by default  and only use it to set HOMEDIR environment variable if populated.  If not set, npm will use the default home directory for the specified user.

### DIFF
--- a/manifests/npm.pp
+++ b/manifests/npm.pp
@@ -8,7 +8,7 @@ define nodejs::npm (
   String $package           = $title,
   $source                   = 'registry',
   Array $uninstall_options  = [],
-  String $home_dir          = "",
+  $home_dir                 = undef,
   $user                     = undef,
   Boolean $use_package_json = false,
 ) {
--- a/manifests/npm.pp
+++ b/manifests/npm.pp
@@ -8,7 +8,7 @@ define nodejs::npm (
   String $package           = $title,
   $source                   = 'registry',
   Array $uninstall_options  = [],
-  $home_dir                 = '/root',
+  String $home_dir          = "",
   $user                     = undef,
   Boolean $use_package_json = false,
 ) {
@@ -74,13 +74,19 @@ define nodejs::npm (
     Nodejs::Npm::Global_config_entry<| title == 'https-proxy' |> -> Exec["npm_install_${name}"]
     Nodejs::Npm::Global_config_entry<| title == 'proxy' |> -> Exec["npm_install_${name}"]
 
+    if !empty($home_dir) {
+      $_homedir = "HOME=${home_dir}"
+    } else {
+      $_homedir = undef
+    }
+
     if $use_package_json {
       exec { "npm_${npm_command}_${name}":
         command     => "${npm_path} ${npm_command} ${options}",
         unless      => $list_command,
         user        => $user,
         cwd         => $target,
-        environment => "HOME=${home_dir}",
+        environment => $_homedir,
         require     => Class['nodejs'],
       }
     } else {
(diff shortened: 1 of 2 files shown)


fnoop:
Hmm, well travis fails because of:
```manifests/npm.pp:11:double_quoted_strings:WARNING:double quoted string containing no variables```

However if you set default to undef (`  String $home_dir          = undef,`) then you get a syntax error:
```
Error: Evaluation Error: Error while evaluating a Resource Statement, Nodejs::Npm[vue install as user mav]: parameter 'home_dir' expects a String value, got Undef (file: /srv/maverick/software/maverick/manifests/maverick-modules/maverick_network/manifests/init.pp, line: 16) on node dom-ubuntu.home
```

The obvious fix is to set $home_dir parameter without a type:
`$home_dir = undef`
But this doesn't follow good practice of setting parameter types..

**Comment by fnoop** [comment]
OK that failing check makes an incorrect assumption, removed

**Comment by fnoop** [comment]
Ooh, a green tick form Travis :)

**Comment by fnoop** [comment]
@ekohl Yes, works fine.

**Comment by juniorsysadmin** [comment]
Hmm, can we go a bit further and add `environment` as a parameter since there may be other environment variables that one may wish to pass onto npm? (See https://docs.npmjs.com/misc/config#environment-variables)

Perhaps something like:
```
Parameters:
$home_dir => undef,
$environment => undef,

if !environment and !home_dir {
  $real_environment = "HOME=${home_dir}"
} else {
  $real_environment = $environment

fnoop:
@juniorsysadmin Not convinced that it's a good idea to encourage passing/using environment variables for npm config.  Better to write manifests or use .npmrc?:
https://docs.npmjs.com/misc/config#npmrc-files

manifests/npm.pp
```diff
@@ -8,7 +8,7 @@
   String $package           = $title,
   $source                   = 'registry',
   Array $uninstall_options  = [],
-  $home_dir                 = '/root',
+  $home_dir                 = undef,
```
igalic: while we're at it, we could just as well give home_dir an `Optional[Stdlib::AbsPath]` type.

manifests/npm.pp
```diff
@@ -74,13 +74,19 @@
     Nodejs::Npm::Global_config_entry<| title == 'https-proxy' |> -> Exec["npm_install_${name}"]
     Nodejs::Npm::Global_config_entry<| title == 'proxy' |> -> Exec["npm_install_${name}"]
 
+    if !empty($home_dir) {
+      $_homedir = "HOME=${home_dir}"
+    } else {
+      $_homedir = undef
+    }
```
igalic: i think a selector is easier to read?
```puppet
$_homedir = $home_dir? {
  undef => undef,
  default => "HOME=${home_dir}",
}
```

spec/defines/nodejs_npm_spec.rb
```diff
@@ -52,10 +52,6 @@
         it 'the exec directory should be /home/npm/packages' do
           is_expected.to contain_exec('npm_install_express').with('cwd' => '/home/npm/packages')
         end
-
-        it 'the environment variable HOME should be /root' do
-          is_expected.to contain_exec('npm_install_express').with('environment' => 'HOME=/root')
-        end
```
Dan33l: Instead of removing the test, why not modify it ?

Dan33l:
Hi @fnoop do need some help to move forward about this PR ?

manifests/npm.pp
ghoneycutt: This changes the default behavior of the module and would likely need a major version bump.

PR #364, opened 2018-05-05
 block creation of File['root_npmrc'] when running on Windows

jhicks-camgian:
File['root_npmrc'] resource causes module to fail on Windows.

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -50,12 +50,14 @@ class nodejs::install {
     }
   }
 
-  file { 'root_npmrc':
-    ensure  => 'file',
-    path    => "${facts['root_home']}/.npmrc",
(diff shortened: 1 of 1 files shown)


juniorsysadmin:
@jhicks-camgian Can you provide further detail - what Puppet version, log output etc?

jhicks-camgian:
Puppet 5.5.1 was used but this is an error that would affect any version when used on Windows.

The code under the condition applied in the pull request assumes a Unix user/group schema with a root user.

When applied on a Windows system it attempts to create a '.npmrc' file in /root owned by uid root/gid 0 (which is not possible on Windows).

juniorsysadmin:
Have created #366 to track a proper fix

PR #367, opened 2018-05-13
Rely on beaker-hostgenerator for docker nodesets

ekohl:
https://github.com/voxpupuli/modulesync_config/pull/453


juniorsysadmin:
> Did #364 break master?

Nah, don't think so.

PR #373, opened 2018-06-01
fix 'nodejs::repo_ensure: absent' for yum

bugfood:
Presently, setting ensure to 'absent' results in creating two files in
/etc/yum.repos.d:

$ cat nodesource.repo
[nodesource]
$ cat nodesource-source.repo
[nodesource-source]

This results in yum failure:
$ sudo yum check-update
Loaded plugins: fastestmirror, security
Repository 'nodesource-source' is missing name in configuration, using id
Repository 'nodesource' is missing name in configuration, using id
Loading mirror speeds from cached hostfile
Error: Cannot find a valid baseurl for repo: nodesource

It is better to ensure these files do not exist. The corresponding apt
class is already set up this way.

Note: I was unable to get local tests working for this module. Please
test before merging.

### DIFF
--- a/manifests/repo/nodesource/yum.pp
+++ b/manifests/repo/nodesource/yum.pp
@@ -58,11 +58,11 @@ class nodejs::repo::nodesource::yum {
   else {
 
     yumrepo { 'nodesource':
-      enabled => 'absent',
+      ensure => 'absent',
     }
 
     yumrepo { 'nodesource-source':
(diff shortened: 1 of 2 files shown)


bugfood:
I had a bit of trouble with the tests at first, but I think I got it straightened out. Please let me know if there is anything that needs fixing.

Thanks,
Corey

PR #386, opened 2019-01-31
Ensure NPM secret keys are idempotent in global_config_entry

### LINKED ISSUE #326: Poor idempotency for NPM "secret" keys

- Puppet: 5
- Ruby: 2.4 (Embedded)
- Distribution: CentOS 7
- Module version: Latest

Use `nodejs::npm::global_config_entry` to declare any "secret" config key in NPM, such as `_auth`, `_password`, or `registry_uri:_password`
## What are you seeing
Puppet is unable to retrieve the existing key due to protections in NPM that prevent it from showing any key that is prefixed with an `_`. This forces it to recreate the resource with every run, preventing a node from ever showing as unchanged. 
## What behaviour did you expect instead
Puppet should be able to verify the key already exists (outside of NPM commands) and is in the desired state without forcing it to be reset with every run
## Any additional information you'd like to impart
PR incoming, looking for feedback on best method to check for these keys. Considering just testing for the value directly against the config, but different OSs have different paths to the global config file and this path is not currently known/managed in the module.

jplindquist:
- Ensure NPM secret keys are idempotent in global_config_entry by matching key/value using npm's regex for secrets (https://github.com/npm/cli/blob/4c65cd952bc8627811735bea76b9b110cc4fc80e/lib/config.js#L179). If a secret is being set, dynamically determine the path to the global config file and match the key and value in the file (outside of npm itself)
- New test(s)

Fixes #326

### DIFF
--- a/manifests/npm/global_config_entry.pp
+++ b/manifests/npm/global_config_entry.pp
@@ -11,17 +11,39 @@ define nodejs::npm::global_config_entry (
 
   case $ensure {
     'absent': {
-      $command        = "config delete ${config_setting} --global"
-      $onlyif_command = $::osfamily ? {
-        'Windows' => "${cmd_exe_path} /C ${npm_path} get --global| FINDSTR /B \"${config_setting}\"",
-        default   => "${npm_path} get --global | /bin/grep -e \"^${config_setting}\"",
+      $command = "config delete ${config_setting} --global"
+
+      # If this is a secret key, determine if it is set properly outside of NPM
+      # https://github.com/voxpupuli/puppet-nodejs/issues/326
+      if $config_setting =~ /(_|:_)/ {
+        $onlyif_command = $::osfamily ? {
+          'Windows' => "${cmd_exe_path} /C FOR /F %G IN ('${npm_path} config get globalconfig') DO IF EXIST %G (FINDSTR /B /C:\"${$config_setting}\" %G) ELSE (EXIT 1)",
+          default   => "test -f $(${npm_path} config get globalconfig) && /bin/grep -qe \"^${$config_setting}\" $(${npm_path} config get globalconfig)",
+        }
+      }
+      else {
+        $onlyif_command = $::osfamily ? {
+          'Windows' => "${cmd_exe_path} /C ${npm_path} get --global| FINDSTR /B \"${config_setting}\"",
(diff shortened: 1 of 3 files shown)


manifests/npm/global_config_entry.pp
```diff
@@ -11,17 +11,39 @@
...
-      $command        = "config set ${config_setting} ${value} --global"
-      $onlyif_command = $::osfamily ? {
-        'Windows' => "${cmd_exe_path} /C FOR /F %i IN ('${npm_path} get ${config_setting} --global') DO IF \"%i\" NEQ \"${value}\" ( EXIT 0 ) ELSE ( EXIT 1 )",
-        default   => "/usr/bin/test \"$(${npm_path} get ${config_setting} --global | /usr/bin/tr -d '\n')\" != \"${value}\"",
+      $command = "config set ${config_setting} ${value} --global"
+
+      # If this is a secret key, determine if it is set properly outside of NPM
+      # https://github.com/voxpupuli/puppet-nodejs/issues/326
+      if $config_setting =~ /(_|:_)/ {
+        $onlyif_command = $::osfamily ? {
```
bastelfreak: could you please replace the legacy topscope varible with the facts hash?
```suggestion
        $onlyif_command = $facts['os']['family'] ? {
```

manifests/npm/global_config_entry.pp
```diff
@@ -11,17 +11,39 @@
...
+          'Windows' => "${cmd_exe_path} /V /C FOR /F %G IN ('${npm_path} config get globalconfig') DO IF EXIST %G (FINDSTR /B /C:\"${$config_setting}=\\\"${$value}\\\"\" %G & IF !ERRORLEVEL! EQU 0 ( EXIT 1 ) ELSE ( EXIT 0 )) ELSE ( EXIT 0 )",
+          default   => "! test -f $(${npm_path} config get globalconfig) || ! /bin/grep -qe '^${$config_setting}=\"${$value}\"' $(${npm_path} config get globalconfig)",
+        }
+      }
+      else {
+        $onlyif_command = $::osfamily ? {
```
bastelfreak:
```suggestion
        $onlyif_command = $facts['os']['family'] ? {
```

bastelfreak:
Thanks!

PR #391, opened 2019-03-25
Avoid collision with other declarations of same resource

AlexanderSalmin:
Avoid collision with other declarations of same resource

Error: Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Evaluation Error: Error while evaluating a Resource Statement, Evaluation Error: Error while evaluating a Function Call, Duplicate declaration

### DIFF
--- a/manifests/repo/nodesource/apt.pp
+++ b/manifests/repo/nodesource/apt.pp
@@ -10,7 +10,7 @@ class nodejs::repo::nodesource::apt {
   include ::apt
 
   if ($ensure != 'absent') {
-    apt::source { 'nodesource':
+    $nodejs_apt_source = {
       include  => {
         'src' => $enable_src,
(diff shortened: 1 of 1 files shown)


bastelfreak:
Hi @AlexanderSalmin, thanks for the PR. I'm not sure if this is a good move. The created repository is pretty module-specific and should not collide. if you manage this already at another place it might make more sense to set `manage_package_repo` to false in this repository? If we merge your PR, we assume that a repo with the same name contains the correct packages, which will lead to other errors.

juniorsysadmin:
What @bastelfreak wrote. @AlexanderSalmin Can you explain more about your particular use case?

juniorsysadmin:
Closing due to no response, feel free to re-open.

Issue #404, opened 2019-10-22
Missing dependency on package resource that installs NPM when using the nodesource repository


wiebe:
## Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: all
- Ruby: all
- Distribution: all
- Module version: probably v0.8.0 and above

## How to reproduce (e.g Puppet code you use)
Try to set a global_config_entry before the nodejs/npm package is being installed. As an example:
```
nodejs::npm::global_config_entry { 'always-auth':
   ensure => 'present',
   value  => 'true',
}
```

## What are you seeing
The npm_config exec fails and the config is not set.

## What behaviour did you expect instead
For the resource to apply without errors.

## Output log
```
Error: /bin/sh: 1: /usr/bin/npm: not found
       
Error: /Nodejs::Npm::Global_config_entry[always-auth]/Exec[npm_config present always-auth]/returns: change from 'notrun' to ['0'] failed: /bin/sh: 1: /usr/bin/npm: not found
```

## Any additional information you'd like to impart
I implemented a simple fix, will make a pull request.

juniorsysadmin:
Fixed by #405

PR #405, opened 2019-10-22
Fix global_config_entry dependency on NPM when using nodesource repo

wiebe:
This fixes the npm_config exec in the global_config_entry class not having a dependency on the package that installs NPM when using the nodesource class.

#404


wiebe:
Not sure why the CI build fails, doesn't seem to be because of my change

bastelfreak:
@wiebe thanks for the PR. Can you take a look at the used email address in the commit? It isn't associated with your github account. Please also add a test for this change.

wiebe:
My email address should be added to this account now. I rebased and also added tests, everything is green now.

Issue #400, opened 2019-08-08
Add Debian 10 (Buster) support


xzavrel:
Hi guys, as Debian Buster has been released it would be nice to have the support for this one also. Is there anything blocking the 

## Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: 6.7.2
- Ruby: 2.5
- Distribution: Debian 10 (Buster)
- Module version: 

## How to reproduce (e.g Puppet code you use)
Use Puppet `class { 'nodejs': }` on Debian 10.

## What are you seeing
Error: /Package[node-red]: Provider npm is not functional on this host

## What behaviour did you expect instead
npm should be installed

mryanb:
Similar issue here.

Dan33l:
@xzavrel and @mryanb hi.
Can you propose a PR ? We will be happy to merge it.

juniorsysadmin:
Closed by #406

PR #403, opened 2019-10-19
Clean up acceptance spec helper

ekohl:

PR #406, opened 2019-10-22
Add Debian 10 support

wiebe:
Adds support for Debian 10

#400


Wiebelt:
The tests fail because of an issue outside of these commits/pull request. See ekohl's comment in #403

wiebe:
I'm not entirely sure how adding acceptance tests would work, add it to the travis yml config i guess but where does the node definition / BEAKER_setfile come from?

ghoneycutt:
The error is a syntax error not related to this pull request

PR #413, opened 2020-01-02
Add support for RHEL 8 based distros

mmoll:

### DIFF
--- a/manifests/repo/nodesource.pp
+++ b/manifests/repo/nodesource.pp
@@ -12,7 +12,7 @@ class nodejs::repo::nodesource {
 
   case $facts['os']['family'] {
     'RedHat': {
-      if $facts['os']['release']['major'] =~ /^[67]$/ {
+      if $facts['os']['release']['major'] =~ /^[678]$/ {
         $dist_version = $facts['os']['release']['major']
         $name_string  = "Enterprise Linux ${dist_version}"
       }
--- a/.sync.yml
+++ b/.sync.yml
@@ -5,6 +5,7 @@
     - set: ubuntu1604-64
     - set: ubuntu1804-64
     - set: centos7-64
+    - set: centos8-64
     - set: debian8-64
     - set: debian9-64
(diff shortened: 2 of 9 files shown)


manifests/repo/nodesource.pp
```diff
@@ -12,7 +12,7 @@
 
   case $facts['os']['family'] {
     'RedHat': {
-      if $facts['os']['release']['major'] =~ /^[67]$/ {
+      if $facts['os']['release']['major'] =~ /^[679]$/ {
```
ghoneycutt: I believe you meant to use `678` and not `9`

ghoneycutt:
Thanks, great looking PR! Could you also add support to the README under https://github.com/voxpupuli/puppet-nodejs#limitations

mmoll:
There are still some problems, at least the one with the `nodejs` package from Nodesource can also be seen on real systems (problem is IMHO in the NodeJS RPM itself), for EPEL, I'll need to have a look...

ghoneycutt:
Appears to trying to load a template for EPEL8 which does not exist.

mmoll:
https://github.com/stahnma/puppet-module-epel/pull/92 is waiting for a new module release.

mmoll:
Fot the nodesource repo to work, `yum module disable -y nodejs` would be needed (see https://github.com/nodesource/distributions/pull/997). Is there a puppet way to do this?

dhoppe:
@mmoll As far as I know ```puppet/yum``` or ```puppetlabs/yumrepo_core``` do not support the handling of modules. 

You could use the template resource to create the following file:
```
[root@1ca0b18b505b ~]# cat /etc/dnf/modules.d/nodejs.module
[nodejs]
name=nodejs
stream=
profiles=
state=disabled
```

mmoll:
@dhoppe Thanks for that hint, seems to work well. :bowing_man: 

This should be ready for review/merge then.

PR #402, opened 2019-08-27
update list of tested OSes during acceptance tests accordingly with metadata.json

Dan33l:

PR #414, opened 2020-01-18
update list of tested OSes accordingly with metadata.json

mmoll:

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -11,7 +11,7 @@ class nodejs::params {
   $repo_proxy_password         = 'absent'
   $repo_proxy_username         = 'absent'
   $repo_release                = undef
-  $repo_url_suffix             = '8.x'
+  $repo_url_suffix             = '12.x'
   $use_flags                   = ['npm', 'snapshot']
 
   $cmd_exe_path = $facts['os']['family'] ? {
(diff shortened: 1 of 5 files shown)


mmoll:
rebased version of GH-402

mmoll:
Debian/buster is already bringing NodeJS 10, so using 8.x as default will not work, as the package is not downgraded, leading to a test fail.

Looking at https://nodejs.org/en/about/releases/ I'd even lean towards setting the default to 12.x.

manifests/params.pp
```diff
@@ -11,7 +11,7 @@
   $repo_proxy_password         = 'absent'
   $repo_proxy_username         = 'absent'
   $repo_release                = undef
-  $repo_url_suffix             = '8.x'
+  $repo_url_suffix             = '10.x'
```
bastelfreak: is this change required? If so it qualifies as a backwards-incompatble change?

mmoll: https://github.com/voxpupuli/puppet-nodejs/pull/414#issuecomment-575895848 :)

bastelfreak: yeah, I'm fine with 12

mmoll: There we go... :green_heart:

juniorsysadmin:
Also consider updating the `repo_url_suffix` section of the README

mmoll:
> Also consider updating the repo_url_suffix section of the README

Done.

PR #425, opened 2020-09-18
Drop EOL Debian 8

bastelfreak:

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -21,7 +21,7 @@ class nodejs::params {
 
   case $facts['os']['family'] {
     'Debian': {
-      if $facts['os']['release']['major'] =~ /^(8|9|10)$/ {
+      if $facts['os']['release']['major'] =~ /^(9|10)$/ {
         $manage_package_repo       = true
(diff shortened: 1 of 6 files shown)


igalic (approved):

manifests/params.pp
```diff
@@ -21,7 +21,7 @@
 
   case $facts['os']['family'] {
     'Debian': {
-      if $facts['os']['release']['major'] =~ /^(8|9|10)$/ {
+      if $facts['os']['release']['major'] =~ /^(9|10)$/ {
         $manage_package_repo       = true
         $nodejs_debug_package_name = 'nodejs-dbg'
```
ekohl, line 26: If I look at https://packages.debian.org/stretch/nodejs-dbg then it looks like this package is not available on Buster, but [npm](https://packages.debian.org/buster/npm) is.

ekohl (approved):
Feel free to fix Debian 10 in another PR.

manifests/params.pp
bastelfreak: I will try to enable acceptance tests for Debian 10 to fix this now and in the future

PR #426, opened 2020-09-18
Drop duplicate class declaration

bastelfreak:
The `contain` not only enforces the ordering, it also adds it to the
catalog. We don't need to call the class explicitly in advance.


igalic (approved):
wow that's so weird i've never seen something like this … "style" before

ekohl:
> wow that's so weird i've never seen something like this … "style" before

It's a valid style and if you need to pass in parameters while also containing it, this is the only way.

However, in this case it's redundant.

PR #431, opened 2020-11-01
Debian and Ubuntu versioning and package name fixes

kenyon:
* Debian has a native nodejs dev package, so define that.

* Debian 10 has a native npm package, so define that for Debian 10.

* Improve `native_debian_devel_package` logic: support Debian 10 onward and Ubuntu 20.04 onward with the libnode-dev package name.

* After the above changes, the variable `is_supported_debian_version` becomes obsolete, so most of the tests can work normally. The only special case is Debian 9 lacking the native npm package.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -22,13 +22,21 @@ class nodejs::params {
   case $facts['os']['family'] {
     'Debian': {
       if $facts['os']['release']['major'] =~ /^(9|10)$/ {
+        $debian_nodejs_dev_package_name = $facts['os']['release']['major'] ? {
+          '9'     => 'nodejs-dev',
+          default => 'libnode-dev',
+        }
+        $debian_npm_package_name = $facts['os']['release']['major'] ? {
+          '9'     => false,
+          default => 'npm',
+        }
         $manage_package_repo       = true
         $nodejs_debug_package_name = 'nodejs-dbg'
-        $nodejs_dev_package_name   = undef
+        $nodejs_dev_package_name   = $debian_nodejs_dev_package_name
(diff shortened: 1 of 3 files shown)


manifests/params.pp
```diff
@@ -22,13 +22,21 @@
   case $facts['os']['family'] {
     'Debian': {
       if $facts['os']['release']['major'] =~ /^(9|10)$/ {
+        $debian_nodejs_dev_package_name = $facts['os']['release']['major'] ? {
+          '9'     => 'nodejs-dev',
```
bastelfreak, line 26: can you add an accetpance test to verify the new settings are correct?

manifests/params.pp
```diff
@@ -22,13 +22,21 @@
...
+          default => 'libnode-dev',
+        }
+        $debian_npm_package_name = $facts['os']['release']['major'] ? {
+          '9'     => false,
```
juniorsysadmin, line 30: I don't like mixing of strings and booleans, can this be done differently?

kenyon: @juniorsysadmin I did it like that to preserve the existing value of `false` for Debian 9. This is also seen in the OpenBSD and Gentoo cases in this file. I suppose we could make it `undef` instead, but that would make it inconsistent with the OpenBSD and Gentoo cases. I could change those also, but that starts to get outside the scope of my pull request.

manifests/params.pp
kenyon: @bastelfreak I added acceptance tests. I'm not sure if the "uninstall" step is the proper way to do this, but it works.

bastelfreak (approved):

PR #432, opened 2020-11-01
add parameter $manage_nodejs_package

kenyon:
Defaults to true. This allows for using this module without managing the nodejs package. The use case for this is where you want to use the npm package provider in this module, but otherwise have no need for directly installing nodejs. This patch allows for expressing that use case in your puppet code.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -1,6 +1,7 @@
 # == Class: nodejs: See README.md for documentation.
 class nodejs (
   $cmd_exe_path                                        = $nodejs::params::cmd_exe_path,
+  Boolean $manage_nodejs_package                       = true,
   Boolean $manage_package_repo                         = $nodejs::params::manage_package_repo,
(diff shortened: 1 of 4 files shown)


kenyon:
Found a problem with this, should be fixed shortly...

kenyon:
Looks good now. :checkered_flag:

manifests/init.pp
```diff
@@ -1,6 +1,7 @@
 # == Class: nodejs: See README.md for documentation.
 class nodejs (
   $cmd_exe_path                                        = $nodejs::params::cmd_exe_path,
+  Boolean $manage_nodejs_package                       = $nodejs::params::manage_nodejs_package,
```
bastelfreak, line 4: `$nodejs::params::manage_nodejs_package` is only used in this singe place. Please directly assign the  value in init.pp and don't add another variable to params.pp. That makes it later on easier to migrate to puppet-strings

kenyon: @bastelfreak Done!

bastelfreak (approved):

PR #435, opened 2020-11-16
fix failures on BSDs by not using fully qualified path

### LINKED ISSUE #434: nodejs::npm::global_config_entry cannot work on a platform that puts npm into a different path

- Puppet: 6.19.1
- Ruby: 2.6.5
- Distribution: FreeBSD
- Module version: 8.1.0

```puppet
  class { 'nodejs':
    nodejs_package_name => 'node12',
    npm_package_name    => 'npm-node12',
    npm_path            => '/usr/local/bin/npm'
  }

  nodejs::npm::global_config_entry { 'unsafe-perm':
    ensure => 'present',
    value  => 'true',
  }
```

## What are you seeing

All of my `nodejs::npm::global_config_entry` are failing, because the code isn't made to run on anything the author didn't anticipate. but it comes down to two things:

```puppet
define nodejs::npm::global_config_entry (
  Enum['present', 'absent'] $ensure = 'present',
  $config_setting                   = $title,
  $cmd_exe_path                     = $nodejs::cmd_exe_path,
  $npm_path                         = $nodejs::params::npm_path,
  $value                            = undef,
) {
  include nodejs
```

`npm_path` defaults to `$nodejs::params::npm_path` rather than `$nodejs::npm_path`, which is wrong.
`$nodejs::params::npm_path` for FreeBSD is also set to `/usr/bin/npm` rather than `/usr/local/bin/npm`.

next, we have a hard-coded path of `/usr/bin/test`, which on FreeBSD is located in `/bin/test`.

Both of these seem… unnecessary. If the `PATH` is correct, we shouldn't have to hard-code path, unless they are *extremely* esoteric.

igalic:

if we can't rely on our PATH, we're hosed anyway.

Fixes #434

### DIFF
--- a/manifests/npm/global_config_entry.pp
+++ b/manifests/npm/global_config_entry.pp
@@ -3,7 +3,7 @@ define nodejs::npm::global_config_entry (
   Enum['present', 'absent'] $ensure = 'present',
   $config_setting                   = $title,
   $cmd_exe_path                     = $nodejs::cmd_exe_path,
-  $npm_path                         = $nodejs::params::npm_path,
+  $npm_path                         = $nodejs::npm_path,
   $value                            = undef,
 ) {
   include nodejs
@@ -17,13 +17,13 @@ define nodejs::npm::global_config_entry (
       if $config_setting =~ /(_|:_)/ {
         $onlyif_command = $facts['os']['family'] ? {
           'Windows' => "${cmd_exe_path} /C FOR /F %G IN ('${npm_path} config get globalconfig') DO IF EXIST %G (FINDSTR /B /C:\"${$config_setting}\" %G) ELSE (EXIT 1)",
-          default   => "test -f $(${npm_path} config get globalconfig) && /bin/grep -qe \"^${$config_setting}\" $(${npm_path} config get globalconfig)",
+          default   => "test -f $(${npm_path} config get globalconfig) && grep -qe \"^${$config_setting}\" $(${npm_path} config get globalconfig)",
         }
       }
       else {
         $onlyif_command = $facts['os']['family'] ? {
           'Windows' => "${cmd_exe_path} /C ${npm_path} get --global| FINDSTR /B \"${config_setting}\"",
-          default   => "${npm_path} get --global | /bin/grep -e \"^${config_setting}\"",
+          default   => "${npm_path} get --global | grep -e \"^${config_setting}\"",
         }
       }
     }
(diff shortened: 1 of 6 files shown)


igalic:
n.b.: I want to add tests to this… for all supported OSes… but OpenBSD & FreeBSD aren't supported right now anyway, so…

manifests/npm/global_config_entry.pp
```diff
@@ -17,13 +17,13 @@
       if $config_setting =~ /(_|:_)/ {
         $onlyif_command = $facts['os']['family'] ? {
           'Windows' => "${cmd_exe_path} /C FOR /F %G IN ('${npm_path} config get globalconfig') DO IF EXIST %G (FINDSTR /B /C:\"${$config_setting}\" %G) ELSE (EXIT 1)",
-          default   => "test -f $(${npm_path} config get globalconfig) && /bin/grep -qe \"^${$config_setting}\" $(${npm_path} config get globalconfig)",
+          default   => "test -f $(${npm_path} config get globalconfig) && grep -qe \"^${$config_setting}\" $(${npm_path} config get globalconfig)",
```
bastelfreak, line 20: if we switch to relative paths, do we need to set `$PATH` in the exec resource?

igalic: :+1:

igalic: fixed

bastelfreak (approved):

spec/defines/global_config_entry_spec.rb
```diff
@@ -11,6 +11,15 @@
         facts
       end
 
+      let(:npm_path) do
+        if os =~ /freebsd/i
```
igalic, line 15:
```suggestion
        if facts[:os]['family'] == 'FreeBSD'
```

spec/defines/nodejs_npm_spec.rb
```diff
@@ -11,6 +11,14 @@
         facts
       end
 
+      let(:npm_path) do
+        if os =~ /freebsd/i
```
igalic, line 15:
```suggestion
        if facts[:os]['family'] == 'FreeBSD'
```

igalic (commented):
let's just directly access the facts object

juniorsysadmin (approved):

spec/defines/global_config_entry_spec.rb
```diff
@@ -11,6 +11,15 @@
...
 
+      let(:npm_path) do
+        if facts[:os]['family'] == 'FreeBSD'
+          '/usr/local/bin/npm'
+        else
+          '/usr/bin/npm'
+        end
+      end
+
+
```
igalic, line 22:
```suggestion
```

igalic (commented):
this line must go to pass rubocop

bastelfreak (approved):

igalic:
this needs a better changelog entry, so we should at least change the title here: we now actually support FreeBSD, and it's metadata.json and as such it now has tests

PR #436, opened 2020-11-19
npm/global_config_entry.pp: improve shell quoting

kenyon:
Improve shell quoting for safety and readability.


manifests/npm/global_config_entry.pp
```diff
@@ -17,13 +17,13 @@
       if $config_setting =~ /(_|:_)/ {
         $onlyif_command = $facts['os']['family'] ? {
           'Windows' => "${cmd_exe_path} /C FOR /F %G IN ('${npm_path} config get globalconfig') DO IF EXIST %G (FINDSTR /B /C:\"${$config_setting}\" %G) ELSE (EXIT 1)",
-          default   => "test -f $(${npm_path} config get globalconfig) && grep -qe \"^${$config_setting}\" $(${npm_path} config get globalconfig)",
+          default   => "test -f \"$(${npm_path} config get globalconfig)\" && grep -qe '^${$config_setting}' \"$(${npm_path} config get globalconfig)\"",
```
igalic, line 20: i don't understand the logic of your choices: when did you pick single quotes and when did you pick double quotes, and, why?

igalic (commented):
🙋🏻‍♀️

manifests/npm/global_config_entry.pp
kenyon: @igalic when we don't need the shell to do any expansion or interpolation, we can use single quotes, which also saves us some escaping of double quotes. Command substitutions should usually be quoted to prevent word splitting and filename expansion on the results. Probably this doesn't matter for the expected values here, but might as well. Also, here
https://github.com/voxpupuli/puppet-nodejs/blob/ed516f61acced781cbb587337a9cf76db1dc5e4d/manifests/npm/global_config_entry.pp#L44
we already had proper quoting of the command substitution, so I wanted to make the other ones consistent.

Shell programming is the worst :wink:

I wonder if this should be using [inifile](https://forge.puppet.com/modules/puppetlabs/inifile) instead, but I guess this works. :man_shrugging:

igalic: actually, using inifile would make this infinitely more readable, and easier to transport from one OS to another, and simplify our testing, greatly, as the platform abstraction would already be handled by inifile itself

igalic: so, what do we do?
who has time for this?

kenyon: (just rebased on current master)

@igalic maybe I can get to a conversion to inifile eventually. I created #439 as a reminder.

PR #437, opened 2020-11-25
modulesync 4.0.0

bastelfreak:
modulesync 4.0.0


ekohl (approved):
Ubuntu 20.04 with Puppet 5 is to be expected.

ekohl:
Looks like CentOS 6 breaks. Given it's going EOL I'd suggest dropping it.

PR #441, opened 2021-03-08
nodejs::npm::global_config_entry: reimplement using ini_setting

### LINKED ISSUE #439: npm/global_config_entry.pp: refactor to use inifile
It may be better to use the inifile module rather than complicated exec resources in `npm/global_config_entry.pp`. See discussion in #436.

kenyon:
Manage npm global config entries using the `ini_setting` resource instead of `exec` resources using `npm config` commands.

Closes #439

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -1,9 +1,10 @@
 fixtures:
   repositories:
-    "stdlib": "git://github.com/puppetlabs/puppetlabs-stdlib.git"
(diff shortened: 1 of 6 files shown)


kenyon:
Seems like something weird about npm in docker: it wants to use `/usr/etc/npmrc` instead of `/etc/npmrc`, but of course `/usr/etc` doesn't exist.

PR #479, opened 2023-09-25
Release 10.0.0

evgeni:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,13 +4,29 @@ All notable changes to this project will be documented in this file.
 Each new release typically also includes the latest modulesync defaults.
 These should not affect the functionality of the module.
 
-## [v9.1.0](https://github.com/voxpupuli/puppet-nodejs/tree/v9.1.0) (2022-01-18)
+## [v10.0.0](https://github.com/voxpupuli/puppet-nodejs/tree/v10.0.0) (2023-09-25)
+
+[Full Changelog](https://github.com/voxpupuli/puppet-nodejs/compare/v9.1.0...v10.0.0)
+
+**Breaking changes:**
+
+- Drop support for Debian 9, Ubuntu 16.04 and 18.04, FreeBSD 11, add support for Puppet 8, add support for RHEL 9 [\#453](https://github.com/voxpupuli/puppet-nodejs/pull/453) ([bastelfreak](https://github.com/bastelfreak))
+
+## [v9.1.0](https://github.com/voxpupuli/puppet-nodejs/tree/v9.1.0) (2023-09-16)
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -4,14 +4,30 @@ All notable changes to this project will be documented in this file.
...
+
+- Drop support for Debian 9, Ubuntu 16.04 and 18.04, FreeBSD 11, add support for Puppet 8, add support for RHEL 9 [\#453](https://github.com/voxpupuli/puppet-nodejs/pull/453) ([bastelfreak](https://github.com/bastelfreak))
+
+## [v9.1.0](https://github.com/voxpupuli/puppet-nodejs/tree/v9.1.0) (2023-09-16)
 
 [Full Changelog](https://github.com/voxpupuli/puppet-nodejs/compare/v9.0.1...v9.1.0)
 
+**Breaking changes:**
+
+- Drop Puppet 6 support [\#467](https://github.com/voxpupuli/puppet-nodejs/pull/467) ([bastelfreak](https://github.com/bastelfreak))
```
evgeni, line 21: Should we yank the 9.1.0 release because of that?

evgeni:
Could also wait for https://github.com/voxpupuli/puppet-nodejs/pull/480 :)

CHANGELOG.md
zilchms: in my opinion this is basically a breaking change und should require a major release. but maybe there are some other opinions (puppet6 is EOL so...)

zilchms:
> Could also wait for #480 :)

Seems like a good idea, shouldnt take too long to get that one merged

evgeni:
> > Could also wait for #480 :)
> 
> Seems like a good idea, shouldnt take too long to get that one merged

So I thought, but then we don't have Debian 12 support in puppet_metadata yet and Ubuntu 22.04 tests seem to timeout for some funny reason I couldn't care about right now.

CHANGELOG.md
evgeni: That's also my reasoning here.

Ideally we would yank that release from Forge and tell GHC to fold the changelog into the 10.0 one, but I have no idea how.

evgeni: GHC has a `exclude-tags` option, but I don't seem to be able to get it to work :cry:

zilchms: on the other hand: whats done is done, yanking it now after the fact may create even more problems for others. i vote for just leaving it the way it is now

zilchms:
then lets leave it out for now, we could throw that in a 10.1.0 release if necessary

zilchms (approved):

PR #480, opened 2023-09-25
Add support for Debian 12

evgeni:

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -44,7 +44,8 @@
       "operatingsystem": "Debian",
       "operatingsystemrelease": [
         "10",
-        "11"
+        "11",
+        "12"
       ]
     },
     {


evgeni:
Why did CI not run on D12?

manifests/params.pp
```diff
@@ -11,7 +11,7 @@
   $repo_proxy_password         = 'absent'
   $repo_proxy_username         = 'absent'
   $repo_release                = undef
-  $repo_url_suffix             = '12.x'
+  $repo_url_suffix             = '18.x'
```
bastelfreak, line 14: Is that a breaking change?

evgeni: (I'll probably split it out to an own PR, once I verified it fixes what I think it should fix on D12/U22)

manifests/params.pp
```diff
@@ -27,7 +27,10 @@
       $nodejs_dev_package_ensure = 'absent'
       $nodejs_package_name       = 'nodejs'
       $npm_package_ensure        = 'absent'
-      $npm_package_name          = 'npm'
+      $npm_package_name          = $nodejs::manage_package_repo ? {
+        true    => false,
+        default => 'npm'
+      }
```
evgeni, line 33: this produces an interesting circular thingy, which seems to accidentally work on p7 but not on p8

however, using `$manage_package_repo` here is useless, as that is not updated when the users sets `nodejs::manage_package_repo`.

HALP!?

bastelfreak (approved):

PR #481, opened 2023-10-27
Update default release to 18.x, EL7 to 16.x

evgeni:

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -11,7 +11,10 @@ class nodejs::params {
   $repo_proxy_password         = 'absent'
   $repo_proxy_username         = 'absent'
   $repo_release                = undef
-  $repo_url_suffix             = '18.x'
+  $repo_url_suffix             = ($facts['os']['family'] == 'RedHat' and $facts['os']['release']['major'] == '7') ? {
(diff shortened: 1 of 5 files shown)


evgeni:
Yes, the latest LTS is 20.x, but that had failures in CI for me and I figured let's at least update to a supported release.

manifests/params.pp
```diff
@@ -11,7 +11,10 @@
   $repo_proxy_password         = 'absent'
   $repo_proxy_username         = 'absent'
   $repo_release                = undef
-  $repo_url_suffix             = '12.x'
+  $repo_url_suffix             = ($facts['os']['family'] == 'RedHat' and $facts['os']['release']['major'] == '7') ? {
```
bastelfreak, line 14: does that render properly in puppet-strings or do you need to write it on a single line?

evgeni: neither.

"Default value: `$nodejs::params::repo_url_suffix`" is what it documents today.

bastelfreak: ah lol that's params.pp

bastelfreak (approved):

PR #483, opened 2023-10-27
Add support for Ubuntu 22.04

evgeni:

### DIFF
--- a/spec/support/acceptance/purge.rb
+++ b/spec/support/acceptance/purge.rb
@@ -1,0 +1,8 @@
+# frozen_string_literal: true
+
+def purge_node
+  case fact('osfamily')
+  when 'Debian'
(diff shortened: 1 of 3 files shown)


spec/support/acceptance/purge.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+def purge_node
+  case fact('osfamily')
+  when 'Debian'
+    on default, 'apt-get purge -y libnode*', { acceptable_exit_codes: [0, 100] }
```
evgeni, line 6: when you purge things, `libnodeXX` remains on the system and conflicts with files from nodesource packages :roll_eyes:

bastelfreak (approved):

PR #486, opened 2023-10-28
only ensure real npm packages

evgeni:
nodesource's `nodejs` packages have `Provides: npm` which means that when we try to `ensure => absent` the `npm` package, it really removes the `nodejs` package we want.

by setting `allow_virtual => false` we ensure that we only remove real packages.

Fixes: 160ea93dbc6f6633d01f0a078541bbad21ad93ba

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -45,9 +45,17 @@ class nodejs::install {
 
   # npm
   if $nodejs::npm_package_name and $nodejs::npm_package_name != false {
+    # the nodesource nodejs packages provide "npm" which makes Puppet
+    # try to uninstall them again
+    if $nodejs::npm_package_ensure == absent and $nodejs::manage_package_repo == true and $nodejs::repo_class == 'nodejs::repo::nodesource' {
+      $allow_virtual = false
+    } else {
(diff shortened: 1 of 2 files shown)


manifests/install.pp
```diff
@@ -46,8 +46,9 @@
   # npm
   if $nodejs::npm_package_name and $nodejs::npm_package_name != false {
     package { $nodejs::npm_package_name:
-      ensure => $nodejs::npm_package_ensure,
-      tag    => 'nodesource_repo',
+      ensure        => $nodejs::npm_package_ensure,
+      allow_virtual => false,
```
evgeni, line 50: I've debated whether this should be `allow_virtual => $nodejs::npm_package_ensure != 'absent'` or even

```puppet
if $nodejs::npm_package_ensure == absent {
  $allow_virtual = false
} else {
  $allow_virtual = undef
}
```

WDYT?

evgeni: Or, even stronger, add `manage_package_repo` to the mix, as we have no friggin idea how other packages might be crazy.

bastelfreak (approved):

PR #487, opened 2023-10-29
replace repo_url_suffix with repo_version

evgeni:
the fact that the version is the suffix of the URL is an implementation detail of the nodesource repository, which shouldn't leak to the consumers of the module

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -23,7 +23,7 @@ class nodejs (
   $repo_proxy_password                                 = $nodejs::params::repo_proxy_password,
   $repo_proxy_username                                 = $nodejs::params::repo_proxy_username,
   Optional[String] $repo_release                       = $nodejs::params::repo_release,
-  $repo_url_suffix                                     = $nodejs::params::repo_url_suffix,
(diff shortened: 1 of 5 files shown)


manifests/init.pp
```diff
@@ -23,7 +23,7 @@
   $repo_proxy_password                                 = $nodejs::params::repo_proxy_password,
   $repo_proxy_username                                 = $nodejs::params::repo_proxy_username,
   Optional[String] $repo_release                       = $nodejs::params::repo_release,
-  $repo_url_suffix                                     = $nodejs::params::repo_url_suffix,
+  $repo_version                                        = $nodejs::params::repo_version,
```
ekohl, line 26: If we're doing breaking changes, perhaps change this to have a data type?
```suggestion
  String[1] $repo_version                              = $nodejs::params::repo_version,
```

evgeni: sure

ekohl (approved):
I'm debating a non-breaking version, but wonder if it's needed.

evgeni:
The next release will be breaking anyway (in terms of SemVer). We could keep the class api compatible, but the change is easy to do.

PR #493, opened 2024-01-03
Test NodeJS 20 on all ELs

evgeni:


evgeni:
Debian failure unrelated, fix in https://github.com/voxpupuli/puppet-nodejs/pull/494

